PR #16, opened 2025-11-21
feat: support OpenAI compatible llm

ylxmf2005:
Add support for custom OpenAI-compatible API via optional `OPENAI_BASE_URL`, default to `https://api.openai.com/v1`
This enables the use of third-party providers that follow the openai api format.


DayuanJiang:
Thank you for your contribution.

Issue #20, opened 2025-11-28
希望所有平台都能配置代理baseUrl


Fugitive844:
<img width="1650" height="866" alt="Image" src="https://github.com/user-attachments/assets/93eb108a-0b22-4b88-99fb-466e7cbc74f4" />

PR #28, opened 2025-11-30
fix: add customize anthropic baseURL

clevertension:
i use some ai integration platform that can use openai, anthropic etc...., so add customize anthropic baseURL is convenient


DayuanJiang:
Merged. Thank you for you contribution.

DayuanJiang:
This also solved #20

PR #82, opened 2025-12-04
fix: clear files when clicking text-only examples

broBinChen:
### Problem
When clicking on the first example card (Animated Diagram) or the fourth example card (Creative Drawing), the file list was not being cleared. This caused previously uploaded images from the second or third example cards (AWS Architecture or Replicate Flowchart) to incorrectly carry over to text-only examples.

### Solution
Added `setFiles([])` call to the `onClick` handlers of both text-only example cards to ensure the file list is cleared when setting text-only inputs.


DayuanJiang:
Thanks. Merged

PR #85, opened 2025-12-04
refactor: replace deprecated addToolResult with addToolOutput

broBinChen:
### Changes
Replaced the deprecated `addToolResult` method with the new `addToolOutput` method from `ai`.

### Reason
The `addToolResult` API has been marked as deprecated. This update ensures compatibility with future versions of the AI SDK and follows the recommended API usage.

### DIFF
--- a/components/chat-panel.tsx
+++ b/components/chat-panel.tsx
@@ -64,7 +64,7 @@ export default function ChatPanel({
     // Generate a unique session ID for Langfuse tracing
     const [sessionId, setSessionId] = useState(() => `session-${Date.now()}-${Math.random().toString(36).slice(2, 9)}`);
 
-    const { messages, sendMessage, addToolResult, status, error, setMessages } =
+    const { messages, sendMessage, addToolOutput, status, error, setMessages } =
         useChat({
             transport: new DefaultChatTransport({
                 api: "/api/chat",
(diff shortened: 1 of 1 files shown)


Finyouknowwhatyouknow:
I am an user, I am sure about what you've changed but it impacts the way AI gen diagram, it always stuck generating code to me

broBinChen:
> I am an user, I am sure about what you've changed but it impacts the way AI gen diagram, it always stuck generating code to me

Thanks for the feedback! However, I'm not sure why this change would affect diagram generation. Looking at the AI SDK source code (line 9307 in index.js), addToolResult is simply an alias:
image

They're the same function at runtime. Could you share more details about what's happening? Any error messages or specific steps to reproduce? This will help me investigate whether it's truly related to this PR or a separate issue.

Finyouknowwhatyouknow:
image

here is when I try to promt the simple query to create a BPMN diagram, it's just keeping load forever and didn't draw anything

broBinChen:
> image
> here is when I try to promt the simple query to create a BPMN diagram, it's just keeping load forever and didn't draw anything

Thanks for reporting! Since addToolResult and addToolOutput are functionally identical, this issue is likely unrelated to this PR. I've tested on my end and diagram generation works fine.
If the problem persists, please open a separate issue with more details. Thanks! 🙏
image

Finyouknowwhatyouknow:
2 days ago it's still good, very good, this is the diagram I have created 2 days ago with the same promt

image

DayuanJiang:
@broBinChen Thanks for contributing!

@Finyouknowwhatyouknow Regarding the hang issue, I've opened an issue and am actively investigating the root cause. It's difficult to debug since the issue is hard to reproduce consistently. If you know a reliable way to reproduce it, please let me know.

Finyouknowwhatyouknow:
Write non-life insurance claiming process for US market by sequence diagram => here is my promt

here is what the app response

1. the flow was drawed messsy 
2. and the code just keep loading like this 

image

PR #134, opened 2025-12-06
fix: handle edge cases in chat panel and message display

DayuanJiang:
## Summary

- Handle undefined `edit.search`/`edit.replace` in EditDiffDisplay to prevent runtime errors
- Handle empty `chartXML` when displaying diagrams (use converted XML directly instead of calling `replaceNodes`)
- Add missing access code header to regenerate and edit message handlers

## Test plan

- [ ] Verify regenerate message works with access code enabled
- [ ] Verify edit message works with access code enabled
- [ ] Verify diagrams can be displayed when chartXML is empty


DayuanJiang:
Superseded by #135 which includes these bug fixes along with the auto-save feature.

PR #163, opened 2025-12-08
feat: add support for self-hosted draw.io via DRAWIO_BASE_URL environment variable

terrydash:
## Summary
Add support for self-hosted draw.io instances via build-time configuration.

## Problem
In some corporate environments, `embed.diagrams.net` is blocked by network policies. 
Users cannot use the application without access to the default draw.io embed URL.

## Solution
- Add `NEXT_PUBLIC_DRAWIO_BASE_URL` environment variable support
- Pass the `baseUrl` prop to the `DrawIoEmbed` component
- Configure Dockerfile to accept build-time argument for the draw.io URL

## Usage
```yaml
# docker-compose.yaml
services:
  drawio:
    image: jgraph/drawio:latest
    ports:
      - "8080:8080"
  
  next-ai-draw-io:
    build:
      context: .
      args:
        - NEXT_PUBLIC_DRAWIO_BASE_URL=http://drawio:8080
```

Or build directly:
```bash
docker build --build-arg NEXT_PUBLIC_DRAWIO_BASE_URL=http://localhost:8080 -t next-ai-draw-io .
```

**Note:** This is a build-time configuration. To change the draw.io URL, you need to rebuild the Docker image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,8 @@ Open [http://localhost:3000](http://localhost:3000) in your browser.
 
 Replace the environment variables with your preferred AI provider configuration. See [Multi-Provider Support](#multi-provider-support) for available options.
 
+> **Corporate Networks:** If `embed.diagrams.net` is blocked, see [Self-Hosted Draw.io](./docs/self-hosted-drawio.md) for configuration options.
+
 ### Installation
 
 1. Clone the repository:
--- a/docs/self-hosted-drawio.md
+++ b/docs/self-hosted-drawio.md
@@ -1,0 +1,57 @@
+# Self-Hosted Draw.io Configuration
+
+In some corporate environments, `embed.diagrams.net` is blocked by network policies. This guide explains how to use a self-hosted draw.io instance with Next AI Draw.io.
+
+## Configuration
+
+Set the `NEXT_PUBLIC_DRAWIO_BASE_URL` environment variable to point to your self-hosted draw.io instance. This is a **build-time** configuration, meaning you need to rebuild the Docker image to change the URL.
+
+## Docker Build
+
+```bash
+# Build with custom draw.io URL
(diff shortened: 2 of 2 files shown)


DayuanJiang:
Thanks for the PR!  

One issue: the current approach causes the draw.io iframe to load twice - first with
the default URL, then again after the API responds. Users will see a flicker.

Simpler fix: use `NEXT_PUBLIC_DRAWIO_BASE_URL` instead. Next.js makes these available
on the client at build time, so no API needed.

In `app/page.tsx`, just add this at the top (outside the component):

```tsx
const drawioBaseUrl =
    process.env.NEXT_PUBLIC_DRAWIO_BASE_URL || "https://embed.diagrams.net"
```
Then pass it to` <DrawIoEmbed baseUrl={drawioBaseUrl} />`

You can delete the `/api/drawio-config endpoint`.

terrydash:
> Thanks for the PR!
> 
> One issue: the current approach causes the draw.io iframe to load twice - first with the default URL, then again after the API responds. Users will see a flicker.
> 
> Simpler fix: use `NEXT_PUBLIC_DRAWIO_BASE_URL` instead. Next.js makes these available on the client at build time, so no API needed.
> 
> In `app/page.tsx`, just add this at the top (outside the component):
> 
> ```tsx
> const drawioBaseUrl =
>     process.env.NEXT_PUBLIC_DRAWIO_BASE_URL || "https://embed.diagrams.net"
> ```
> 
> Then pass it to` <DrawIoEmbed baseUrl={drawioBaseUrl} />`
> 
> You can delete the `/api/drawio-config endpoint`.

Thanks, I removed it and test ok in my local.

DayuanJiang:
Thanks for the fix!
I pushed a small commit to add documentation - created docs/self-hosted-drawio.md with setup instructions and added a reference in the README.
LGTM otherwise.

Issue #210, opened 2025-12-11
[Bug] 恢复 Chat 窗口后会清空更改


Higurashi-kagome:
复现步骤：

1. 通过 Chat 生成一幅图之后，关闭 Chat 面板
   
    <img width="815" height="171" alt="Image" src="https://github.com/user-attachments/assets/3228884b-d481-47ed-b737-f00d18e8e98a" />

2. 手动修改这幅图
3. 打开面板，手动修改将会丢失（还原到了一开始通过 Chat 生成的图）

    <img width="631" height="227" alt="Image" src="https://github.com/user-attachments/assets/375a80d5-49bb-4bf9-a67a-a658a3788c76" />

正常情况下，应该保留更改。

DayuanJiang:
谢谢反馈。Edgecase测试的时候没考虑。我会尽快更改。

Biki-dev:
@Higurashi-kagome Could you please send a video showing how to reproduce this issue? In my case, I’m not seeing any problem.

Higurashi-kagome:
This is my case🤔:

<details>
<summary>Details</summary>

![Image](https://github.com/user-attachments/assets/8fde9f92-2479-46c0-8c80-07cc44ec454a)

</details>

@Biki-dev

Biki-dev:
> Details

got it ..

DayuanJiang:
Solved.

PR #224, opened 2025-12-11
Persist processed tool calls to prevent replay after chat restore

Biki-dev:
## Summary
- persist processed tool call IDs at the chat panel level
- pass the set into the message display to avoid reprocessing on collapse/restore
- keep manual edits intact when reopening the chat window

issue #210


DayuanJiang:
Thanks. Neat solution. Merged.

Issue #226, opened 2025-12-11
[Bug] custom model setting bug


Twelveeee:
## Bug Description
When a user inputs a custom model, after that selects Use Sever Defalut model, the application incorrectly send `x-ai-api-key` parameter  to the backend. This causes a backend validation or processing error, resulting in an exception or failed request.

## Screenshots
<img width="1653" height="1019" alt="Image" src="https://github.com/user-attachments/assets/add5fb00-e4f7-4616-b7fa-bb357cca7dff" />

PR #227, opened 2025-12-11
fix:custom model setting bug

Twelveeee:
Fixed https://github.com/DayuanJiang/next-ai-draw-io/issues/226

When the Provider is selected as Usu Server Default, the request will no longer pass x-ai-base-url, x-ai-api-key, x-ai-model


DayuanJiang:
Thanks for you contribution! Merged.

PR #229, opened 2025-12-11
feat: Add a new chat button with a confirmation modal

shashikiran797:


DayuanJiang:
Thanks for the contribution! 

I pushed a small fix that addresses the Copilot comments - switched to handleFileChange([]) to properly clear pdfData, and added try-catch for localStorage operations.

PR #248, opened 2025-12-13
feat: add warning dialog for theme and UI style changes

### LINKED ISSUE #243: [Bug] All changes are lost when switching the theme
#### **Description**
When the theme is changed, all existing edits (shapes, images, etc.) are removed automatically. This results in loss of user work without any warning.

---

### **Steps to Reproduce**
1. Add any shape or make any edits.  
2. Open **Settings**.  
3. Change the **Theme**.  

---

### **Actual Behavior**
- The theme changes, but all previously added shapes, images, and edits are removed automatically.

### **Expected Behavior**
- Changing the theme should **not delete user edits**, or  
- The app should **warn the user** that unsaved edits will be lost before applying the theme change.

---

### **Possible Solutions**
1. **Show an alert** before changing the theme, informing the user that unsaved edits will be lost and allowing them to cancel or save first.
2. **Automatically save the current state** before applying the new theme so that user edits remain intact.

---

### **Additional Note**
If the maintainers agree, I would like permission to work on this issue.

Biki-dev:
### 📝 Description
This PR addresses the issue where changing themes (dark/light mode) or draw.io UI styles (min/sketch) would automatically remove all user edits (shapes, images, etc.) without warning. The solution implements a warning system to preserve user work.

### 🎯 Problem Statement
- Changing themes or UI styles caused the DrawIoEmbed component to remount, losing all unsaved edits
- Users lost their work without any warning or opportunity to save
- Poor user experience with potential data loss

### ✅ Solution Implemented
1. **Warning System**: Shows a toast notification when user attempts to change theme/UI style
2. **User Choice**: Provides "Continue" and "Cancel" options with clear warnings

### 🔗 Related Issues
Fixes #243

### DIFF
--- a/contexts/diagram-context.tsx
+++ b/contexts/diagram-context.tsx
@@ -94,10 +94,7 @@ export function DiagramProvider({ children }: { children: React.ReactNode }) {
                     drawioRef.current?.exportDiagram({ format: "xmlsvg" })
                 }),
                 new Promise<string>((_, reject) =>
-                    setTimeout(
-                        () => reject(new Error("Export timeout")),
-                        10000,
-                    ),
+                    setTimeout(() => reject(new Error("Export timeout")), 2000),
                 ),
             ])
 


DayuanJiang:
@Biki-dev  Thanks for the PR! I think it would be better to save the user's modifications directly rather than showing a warning dialog.

Biki-dev:
> @Biki-dev Thanks for the PR! I think it would be better to save the user's modifications directly rather than showing a warning dialog.

ohk ..

Biki-dev:
@DayuanJiang done

DayuanJiang:
Hey @Biki-dev, thanks for tackling this issue!

The approach looks right, but I noticed a couple things:
1. The Promise can hang forever
If DrawIO fails to respond, there's no timeout so the UI just freezes. The codebase already handles this in chat-panel.tsx - check out `onFetchChart` around line 111, it uses Promise.race with a 2 second timeout.
2. `handleDarkModeChange` and `handleDrawioUiChange` are almost identical
You could extract the save logic into a helper like `saveDiagramBeforeChange()` and call it from both handlers. Less duplication, easier to maintain.
3. Minor: the toast might be confusing
"diagram preserved" shows even if the save failed or was skipped. Maybe just remove it? Users don't really need confirmation for every theme toggle.

Biki-dev:
> 1. 2 second timeout.

In the chat-panel, I can see a 10-second timeout.

DayuanJiang:
Good job, merged.

PR #274, opened 2025-12-16
feat: add Vercel AI Gateway support

tedcao:
- Updated environment configuration to include AI_GATEWAY_API_KEY for unified access to multiple AI providers.
- Added gateway provider to the list of supported AI providers in the codebase.
- Enhanced documentation to explain the usage of Vercel AI Gateway and its model format.

This change simplifies authentication and allows users to switch between providers seamlessly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,19 @@
 {
     "name": "next-ai-draw-io",
-    "version": "0.4.0",
+    "version": "0.4.2",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "next-ai-draw-io",
-            "version": "0.4.0",
+            "version": "0.4.2",
             "license": "Apache-2.0",
             "dependencies": {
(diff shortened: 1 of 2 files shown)


DayuanJiang:
Thanks a lot for the contribution. LGTM.
Could update these two minor things?
1. Update package version to latest (@ai-sdk/gateway@^2.0.21 or use the latest)
2. Consider using the standard env variable name `VERCEL_AI_GATEWAY_API_KEY` instead of `AI_GATEWAY_API_KEY` for consistency with the SDK's default behavior.

tedcao:
Packaged upgrade to latest version.
Based on the official doc: https://vercel.com/docs/ai-gateway/getting-started
the SDK default env name is: AI_GATEWAY_API_KEY; I would not suggest to update it.

DayuanJiang:
@tedcao You're right, thanks for catching that.
I've tested it locally and it works. Merged.

PR #279, opened 2025-12-16
feat: Stdio MCP MVP

uetuluk:
Added a stdio MCP prototype.

PNG export is currently broken, but the remaining functionality works as expected. Even in its current state, it already helps agents understand how to construct valid draw.io graphs. Though, further testing is needed to compare agent behavior with and without this MCP enabled.

Built following the Vercel MCP school of thought: most “tools” don’t actually do anything — they just "politely" tell the model “this isn’t a tool, but here’s how you should do the thing.”


DayuanJiang:
Hi, @uetuluk 

First, I want to say a huge thank you for this contribution. I was genuinely impressed when I saw the scale of this PR—7,000+ lines of code is a massive amount of work, and it clearly shows your dedication and enthusiasm for the project.

I have been working on an MCP implementation that takes a different approach. My core vision for this project is **two-way real-time collaboration**, where users can edit diagrams in the browser while the AI updates them simultaneously. Your implementation focuses on stdio transport for Claude Desktop (one-way generation). While this is valid, I am building an HTTP-based MCP integration that connects directly with the browser session to enable that interactive, two-way editing experience. 

Because our architectural paths diverge here, I won't be able to merge this.

For the future, I would kindly suggest that for significant features or architectural additions, it's best to open an Issue first to discuss the approach. That way, we can align before you invest so much time in coding. I would have loved to catch this earlier so we could have collaborated on the HTTP approach instead.

I'm closing this PR, but please don't let this discourage you. 
Some of your ideas are very valuable, like pre-built XML templates for common diagram types, or style property documentation. If you're interested in helping with the HTTP-based MCP integration,please let me know.

Thank you again for your hard work and understanding.

PR #286, opened 2025-12-17
 feat: Add support for sglang provider and improve streaming UX

pointerhacker:
Summary

  This PR introduces support for custom sglang deployments that are OpenAI-compatible
  but have minor deviations from the official API specification. It also improves the
  frontend user experience by adding a loading indicator for multi-step AI responses.

  This resolves issues where the application would fail to display results or appear
  unresponsive when connected to a custom sglang service.

  Problem

  Connecting the application to a custom sglang service revealed three core issues:

   1. Stream Validation Errors: The service's streaming response included a delta object
      with role: "", which is not compliant with the Vercel AI SDK's parser, causing
      AI_TypeValidationError.
   2. Silent Stream Failure: The delta object also contained a non-standard
      reasoning_content: "" field. This caused the AI SDK's parser to silently discard any
      data chunks containing tool calls, preventing diagrams from being rendered.
   3. Poor User Experience: Even when the stream was partially working, there was a long
      pause between the AI's initial text response and the start of the diagram generation,
      making the application feel unresponsive.

  Solution

  This PR implements a robust, multi-part solution:

   1. Added a `sglang` AI Provider (`lib/ai-providers.ts`):
       * A new sglang provider option has been added.
       * It uses the createOpenAI factory but with a custom fetch wrapper to intercept
         and sanitize the response stream.

   2. Implemented a Stream Transformer (`lib/ai-providers.ts`):
       * The custom fetch wrapper pipes the response from the sglang service through a
         TransformStream.
       * This transformer correctly parses Server-Sent Events (SSE) and performs two
         cleanup actions on each data chunk before forwarding it to the AI SDK:
           * Deletes the invalid role: "" from the delta object.
           * Deletes the non-standard reasoning_content field from the delta object.
       * This ensures the AI SDK receives a clean, fully compliant data stream, resolving
         all parsing and rendering errors.

   3. Improved Frontend Feedback (`components/chat-message-display.tsx`):
       * A loading indicator (Generating diagram... with a spinner) has been added to the
         chat display.
       * This indicator now appears whenever the chat is waiting for the assistant to
         continue its response (e.g., during the pause between text and tool generation),
         providing clear feedback to the user that the AI is still working. This is a
         general UX improvement for all providers.

  How to Test

   1. Create or update your .env.local file with the following variables:

   1     AI_PROVIDER=sglang
   2     AI_MODEL=<your-sglang-model-name>
   3     SGLANG_API_KEY=<your-sglang-api-key>
   4     SGLANG_BASE_URL=<your-sglang-base-url>

   2. Run the application and start a chat.
   3. Ask the AI to draw a diagram.
   4. Verify that:
       * The initial text response appears.
       * The "Generating diagram..." indicator appears immediately after.
       * The diagram is successfully drawn on the canvas.
       * No errors appear in the browser or server console.

### DIFF
--- a/docs/ai-providers.md
+++ b/docs/ai-providers.md
@@ -136,6 +136,23 @@ Optional custom URL:
 OLLAMA_BASE_URL=http://localhost:11434
 ```
 
+### Vercel AI Gateway
+
+Vercel AI Gateway provides unified access to multiple AI providers through a single API key. This simplifies authentication and allows you to switch between providers without managing multiple API keys.
+
+```bash
+AI_GATEWAY_API_KEY=your_gateway_api_key
+AI_MODEL=openai/gpt-4o
+```
+
+Model format uses `provider/model` syntax:
+
+-   `openai/gpt-4o` - OpenAI GPT-4o
+-   `anthropic/claude-sonnet-4-5` - Anthropic Claude Sonnet 4.5
+-   `google/gemini-2.0-flash` - Google Gemini 2.0 Flash
+
+Get your API key from the [Vercel AI Gateway dashboard](https://vercel.com/ai-gateway).
+
 ## Auto-Detection
 
 If you only configure **one** provider's API key, the system will automatically detect and use that provider. No need to set `AI_PROVIDER`.
@@ -143,7 +160,7 @@ If you only configure **one** provider's API key, the system will automatically 
 If you configure **multiple** API keys, you must explicitly set `AI_PROVIDER`:
 
 ```bash
-AI_PROVIDER=google  # or: openai, anthropic, deepseek, siliconflow, azure, bedrock, openrouter, ollama
+AI_PROVIDER=google  # or: openai, anthropic, deepseek, siliconflow, azure, bedrock, openrouter, ollama, gateway
 ```
 
 ## Model Capability Requirements
--- a/env.example
+++ b/env.example
@@ -1,6 +1,6 @@
 # AI Provider Configuration
 # AI_PROVIDER: Which provider to use
-# Options: bedrock, openai, anthropic, google, azure, ollama, openrouter, deepseek, siliconflow
+# Options: bedrock, openai, anthropic, google, azure, ollama, openrouter, deepseek, siliconflow, gateway
 # Default: bedrock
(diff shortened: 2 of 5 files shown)


DayuanJiang:
Thanks for the contribution, I will check it later.

app/api/chat/route.ts
```diff
@@ -139,6 +139,54 @@ function createCachedStreamResponse(xml: string): Response {
     return createUIMessageStreamResponse({ stream })
 }
 
+const tools = {
```
DayuanJiang, line 142: Is there any reason to move the place of the `tools`?

components/chat-message-display.tsx
```diff
@@ -1355,6 +1358,20 @@ export function ChatMessageDisplay({
                             </div>
                         )
                     })}
+
+                    {/* UX Improvement: Show a "Generating diagram..." indicator during streaming pauses */}
+                    {isLastMessageFromAssistant && status === 'streaming' && (
```
DayuanJiang, line 1363: The loading indicator shows whenever status === 'streaming' and last message is from assistant. 
Please check the gif I posted
PixPin_2025-12-17_15-43-02

DayuanJiang:
@pointerhacker  Thanks for the contribution!
Please check the comment that I posted.

Also, I think it would be better to split this into two PRs:
1. One for the sglang provider support
2. Another for the streaming UX improvement (the \"Generating diagram...\" indicator)

What do you think?

pointerhacker:
Okay, I will split it into two branches.

app/api/chat/route.ts
pointerhacker: There was no specific reason to move the tools object. I've moved it back to its
  original location within the streamText call in the new branch
  feature/sglang-provider

components/chat-message-display.tsx
pointerhacker: You were right, the loading indicator's condition was too broad. I've fixed this in
  the fix/streaming-ux-indicator branch. The 'Generating diagram...' indicator will now
  only appear when the assistant is streaming but has not yet sent any text content.
  This should resolve the issue shown in the GIF.

Issue #305, opened 2025-12-18
[Bug] DrawIO resets and loses unsaved content when window is resized across 768px breakpoint


E66Crisp:
> **Note**: This template is just a guide. Feel free to ignore the format entirely - any feedback is welcome! Don't let the template stop you from sharing your thoughts.

## Bug Description
When manually resizing the browser window across the 768px breakpoint (switching between mobile and desktop layouts), the `ResizablePanelGroup` component updates/remounts, which causes the DrawIO editor to reinitialize. This results in:

1. DrawIO being re-imported/reloaded
2. Any unsaved diagram content being reset/lost
3. Loss of current editing state

## Steps to Reproduce
1. Open the application and create/edit a diagram
2. Make some changes without saving
3. Resize the browser window to cross the 768px width threshold (e.g., from 800px to 750px, or vice versa)
4. Observe that the diagram resets and unsaved changes are lost

## Expected Behavior
The diagram content and editing state should persist when resizing the window across the mobile/desktop breakpoint.

## Actual Behavior
The DrawIO editor remounts and all unsaved changes are lost.

## Additional Context
Any other information about the problem.

E66Crisp:
Hi there, I've made some modifications to address this issue. I hope these changes will be helpful.

PR #306, opened 2025-12-18
fix: Prevent DrawIO remount and data loss when resizing window across 768px breakpoint

E66Crisp:
### Changes Made

I've attempted to fix this issue #305  with the following modifications:

#### 1. Prevent DrawIO Remounting (app/page.tsx)
- **Removed** `key={isMobile ? "mobile" : "desktop"}` from `ResizablePanelGroup`
- **Moved** the `key` prop to the ChatPanel's `ResizablePanel` instead
- This prevents the entire component tree (including DrawIO) from remounting when crossing the 768px breakpoint

#### 2. Refactor Diagram State Management (diagram-context.tsx & chat-panel.tsx)
- **Moved** diagram restoration and auto-save logic from `chat-panel.tsx` to `diagram-context.tsx`

#### 3. Preserve User Input (chat-panel.tsx)
- **Added** sessionStorage persistence for chat input
- User's typing is now preserved even if ChatPanel remounts due to layout changes
- Input is restored on mount and cleared on submit or reset
Fixes: #305

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -104,19 +104,24 @@ export default function Home() {
     }
 
     // Check mobile - save diagram and reset draw.io before crossing breakpoint
+    const isInitialRenderRef = useRef(true)
     useEffect(() => {
         const checkMobile = () => {
             const newIsMobile = window.innerWidth < 768
-            // If crossing the breakpoint, save diagram and reset draw.io
-            if (newIsMobile !== isMobileRef.current) {
+            // If crossing the breakpoint (not initial render), save diagram and reset draw.io
+            if (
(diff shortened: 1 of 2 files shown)


DayuanJiang:
Thanks for the PR! 

I made a few tweaks on top of your changes:

- Added `isInitialRenderRef` to avoid unnecessary save calls when the page first loads (was firing `saveDiagramToStorage` on initial render if user opened on mobile)
- Removed the console.log statements
- Added an eslint-disable comment for the `loadDiagram` dependency since it uses refs internally

Merged

Issue #308, opened 2025-12-18
[Bug] Send and sidebar buttons become inaccessible when chat-panel is resized


E66Crisp:
> **Note**: This template is just a guide. Feel free to ignore the format entirely - any feedback is welcome! Don't let the template stop you from sharing your thoughts.

## Bug Description
When I shrink the chat-panel area, the Send button and sidebar collapse button move outside the visible screen area, making them inaccessible and preventing user interaction.

## Expected Behavior
The sidebar collapse button and Send button should remain accessible and visible under all viewport sizes, ensuring users can always send messages and toggle the sidebar.

## Screenshots
<img width="925" height="741" alt="Image" src="https://github.com/user-attachments/assets/54dafa49-12bd-46e2-b7e2-16917eabafc0" />

## Additional Context
Any other information about the problem.

PR #309, opened 2025-12-18
fix(components): Send and sidebar buttons become inaccessible when chat-panel is resized

E66Crisp:
When I shrink the chat-panel area, the Send button and sidebar collapse button move outside the visible screen area, making them inaccessible and preventing user interaction

fixes: #308


DayuanJiang:
Thank you for you contribution.

PR #320, opened 2025-12-18
feat: add Structurizr DSL export for C4 diagrams (#264)

### LINKED ISSUE #264: [Feature] DSL export to structurizr
I think this feature will be awesome for C4 Diagrams

godhaniripal:
- Add convertToStructurizrDsl utility with smart C4 element detection
- Create export dialog with copy and download functionality
- Integrate export button in chat toolbar
- Support automatic hierarchy detection for systems and containers
- Add README documentation for new feature

Resolves #264


DayuanJiang:
Hey @godhaniripal, thanks for the work on this!                                            
                                                                                             
After thinking it through, I don't think this fits what the app should be doing. Export format conversion is really draw.io's responsibility - we're just wrapping their editor with AI capabilities. 
If we start maintaining our own converters, we're taking on work that belongs upstream and will be a constant maintenance headache(As you can see the multiple inline comment from Github Copilot).
                                                                                           
Appreciate the effort though!

PR #353, opened 2025-12-22
Fix: Localize TPM rate-limit toast via i18n

Biki-dev:
## Fix: Localize TPM rate-limit toast via i18n

Replaced the hardcoded English TPM rate-limit toast with a fully localized i18n message.

### What changed
- Updated `useQuotaManager` to use `useDictionary` + `formatMessage`
- Added translations for **EN / ZH / JA**
- Message now supports `{limit}` and `{seconds}` placeholders

### Result
The TPM rate-limit toast is now properly localized and consistent with the app’s i18n pattern.


DayuanJiang:
Thanks! Fixed some translation issues. Merged.

PR #395, opened 2025-12-24
fix: add scrollable model list in AI Model Configuration dialog

Biki-dev:
Fixes the model configuration dialog where the model list was not scrollable.

solves #393 

issue detail : https://github.com/DayuanJiang/next-ai-draw-io/issues/393#issuecomment-3688934441


DayuanJiang:
Neat fix, thanks.

PR #396, opened 2025-12-24
Hide scrollbar in model selector dropdown while maintaining scroll functionality

Biki-dev:
Before : 
Screenshot 2025-12-24 142709
After : 
Screenshot 2025-12-24 142744


DayuanJiang:
Thanks for the PR! I added a small tweak on top of your changes - instead of just hiding the scrollbar, I also added a gradient shadow at the bottom so users can still tell there's more content to scroll.

Issue #437, opened 2025-12-27
[Enhancement] Improve Chat Panel UI Layout


DayuanJiang:
## Current Behavior

The chat input area currently displays several controls in a single row:
- History button
- Download button
- Image upload button
- Model selector (e.g., "gpt-5-mini")

When the chat panel width is narrow, the icons on the left side get hidden, reducing functionality access.

### 1. Move utility buttons to Settings page(see screenshot 1,2,3)
Move the following controls from the chat input bar to the Settings page:
- History button
- Download button
- (Optionally other less frequently used controls)

This declutters the main chat interface.

### 2. Responsive model selector(see screenshot 4)
When the chat panel width is narrow:
- Hide the model name text and show only the model icon
- This frees up space so important icons remain visible

## Motivation

- **Cleaner interface**: Moving rarely-used controls to Settings reduces visual clutter
- **Better mobile/narrow view experience**: Users can still access all icons without them being hidden
- **Improved usability**: Primary actions (type message, send) are more prominent

## Screenshots

<img width="1296" height="1614" alt="Image" src="https://github.com/user-attachments/assets/ac1bbe17-e6ac-4ea6-a62c-41c5c2c62a69" />

---

This should be split into two PRs:
- [ ] Move utility buttons to Settings page
- [ ] Responsive model selector

If anyone is willing to contribute to this, please comment below.

Biki-dev:
Looks like no one’s active on this, so I’ll take it.

PR #443, opened 2025-12-27
fix: make model selector label responsive to panel width

Biki-dev:
## Changes
- Added responsive behavior to the model selector based on chat panel width.
- Hide the model name and show only the model icon in narrow layouts.
- Automatically restore the model name when the panel width increases.
- Prevent resize feedback loops by observing the container instead of the button.
- Added hysteresis thresholds to avoid flicker during panel resizing.
 issue #437

### DIFF
--- a/components/model-selector.tsx
+++ b/components/model-selector.tsx
@@ -116,8 +116,8 @@ export function ModelSelector({
     const [showLabel, setShowLabel] = useState(true)
 
     // Threshold (px) under which we hide the label (tweak as needed)
-    const HIDE_THRESHOLD = 250
-    const SHOW_THRESHOLD = 250
+    const HIDE_THRESHOLD = 240
+    const SHOW_THRESHOLD = 260
     useEffect(() => {
         const el = wrapperRef.current
         if (!el) return


components/model-selector.tsx
```diff
@@ -112,134 +112,189 @@ export function ModelSelector({
         ? `${selectedModel.modelId} ${dict.modelConfig.clickToChange}`
         : `${dict.modelConfig.usingServerDefault} ${dict.modelConfig.clickToChange}`
 
+    const wrapperRef = useRef<HTMLDivElement | null>(null)
+    const [showLabel, setShowLabel] = useState(true)
+
+    // Threshold (px) under which we hide the label (tweak as needed)
+    const HIDE_THRESHOLD = 250
+    const SHOW_THRESHOLD = 250
```
DayuanJiang: The HIDE_THRESHOLD and SHOW_THRESHOLD are both 250, which defeats the hysteresis mechanism. At exactly 250px width, the state will toggle infinitely

```suggestion
    const HIDE_THRESHOLD = 240
    const SHOW_THRESHOLD = 260
```

Biki-dev:
@DayuanJiang yeah , ig now good

PR #446, opened 2025-12-28
feat: optimize get_shape_library with LLM-based shape filtering

### LINKED ISSUE #418: [Feature] Optimize `get_shape_library` tool with LLM-based shape filtering
> **Note**: This template is just a guide. Feel free to ignore the format entirely - any feedback is welcome! Don't let the template stop you from sharing your ideas.

## Feature Description
Add an LLM call inside the `get_shape_library` tool to filter shapes based on user's request, reducing context window usage from ~22KB to ~2KB for large libraries.

## Problem Context
The `get_shape_library` tool currently returns the **entire** shape library documentation, which consumes significant context window:

| Library | Lines | Size | Shapes |
|---------|-------|------|--------|
| aws4 | 1049 | 22KB | 1032 |
| azure2 | 431 | 11KB | ~400 |
| gcp2 | 315 | 6KB | 298 |
| alibaba_cloud | 328 | 8KB | ~300 |

When a user asks for "draw an AWS Lambda architecture", they only need ~10-20 relevant shapes, not all 1032.

## Proposed Solution
1. Read the full documentation file
2. Parse it to separate **header/usage section** from **shapes list**
3. If shape count > 100, call LLM to pick ~15-20 relevant shapes based on user's input
4. Return: header + usage example (extracted from doc) + filtered shape names only

### Key Requirements
- Use the **same model** already configured for the chat request
- LLM should only return **shape names** - no generation of usage examples
- Usage examples are extracted from the existing doc header
- Small libraries (< 100 shapes) should be returned as-is without LLM call
- Use `generateText` from AI SDK for the internal call

### Implementation Location
- File: `app/api/chat/route.ts`
- Tool: `get_shape_library` (lines 661-722)

## Alternatives Considered
- Hardcoding commonly used shapes per library (less flexible)
- Using a separate smaller model for filtering (more complex setup)

### Benefits
- **Context reduction**: ~22KB → ~2KB for large libraries
- **Better relevance**: AI picks shapes matching user intent
- **Backward compatible**: Small docs unchanged

---

If you want to contribute this feature, please comment below.

vansh-nagar:
Fixes #418
Implements LLM-based filtering to reduce shape library context size for large libraries.

### DIFF
--- a/app/api/chat/route.ts
+++ b/app/api/chat/route.ts
@@ -3,6 +3,7 @@ import {
     convertToModelMessages,
     createUIMessageStream,
     createUIMessageStreamResponse,
+    generateText,
     InvalidToolInputError,
     LoadAPIKeyError,
     stepCountIs,
@@ -184,6 +185,8 @@ async function handleChatRequest(req: Request): Promise<Response> {
         .find((m: any) => m.role === "user")
     const userInputText =
         lastUserMessage?.parts?.find((p: any) => p.type === "text")?.text || ""
+
+    ;(globalThis as any).lastUserInputText = userInputText
 
     // Update Langfuse trace with input, session, and user
     setTraceInput({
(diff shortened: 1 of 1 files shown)


DayuanJiang:
Hey, thanks for tackling this!

I found one big issue - using `globalThis` to pass the user prompt is going to cause problems in production. Since it's shared across all requests, if two users hit the API at the same time, one user's prompt could end up being used for the other's shape filtering. 

The good news is AI SDK 6 already gives you `messages` in the tool execute function, so you can just do:

```typescript
execute: async ({ library }, { messages }) => {
    const lastUserMessage = [...messages].reverse().find(m => m.role === "user")
    const userPrompt = lastUserMessage?.content?.find((p: any) => p.type === "text")?.text || ""
    // ...
}
```

No global state needed.

A few other things:

1. If generateText fails or parsing returns nothing, the user gets zero shapes back (worse than the full list). Wrap it in try/catch and fall back to returning the full content.
2. You might want to check that the shapes the LLM returns actually exist in the original list - it could hallucinate names.
3. AI SDK  has Output.object() with Zod schemas which handles JSON parsing + validation automatically, might be cleaner than manual JSON.parse with regex fallback.

Let me know if you want help with any of this!

gthieleb:
Hi @vansh-nagar @DayuanJiang — I've opened #925, which rebases this PR's work onto current main (this branch was 306 commits behind) and implements the review feedback from above:

- **No more `globalThis`** — the user prompt now comes from the tool execute `messages` option, so no cross-request state
- **Full-library fallback** on any filtering failure — never zero shapes
- **Hallucination guard** — LLM-selected names are validated against the real shape list
- **`Output.object` + Zod** instead of manual `JSON.parse`/regex

All credit for the original implementation goes to @vansh-nagar (commit is preserved with original authorship in #925). Happy to close #925 in favor of this one if @vansh-nagar prefers to pull the changes in here instead.

Issue #458, opened 2025-12-30
[Bug] Root path and `/about` route return 404 after removing proxy.ts in #448


broBinChen:
## Bug Description

After the removal of `proxy.ts` in PR #448, the application returns 404 errors when accessing the root path (`/`) or `/about` route. The proxy file previously handled locale detection and automatic redirection to locale-prefixed routes (e.g., `/en/`, `/zh/`), but this functionality was lost after its removal.

## Steps to Reproduce

1. Open the application in a browser
2. Navigate to the root URL `/`
3. Observe a 404 error page
4. Navigate to `/about`
5. Observe another 404 error page

**Note**: Direct access to locale-prefixed routes like `/en/` or `/zh/about` still works correctly.

## Expected Behavior

When accessing the root path or routes without a locale prefix:
- The application should detect the user's preferred locale from:
  - Browser language settings
  - Previously saved locale in localStorage (`next-ai-draw-io-locale`)
  - Default to `en` if no preference is found
- Automatically redirect to the appropriate locale-prefixed path:
  - `/` → `/${locale}/` (e.g., `/en/`, `/zh/`, `/ja/`)
  - `/about` → `/${locale}/about`

## Actual Behavior

- Accessing `/` returns a 404 error page
- Accessing `/about` returns a 404 error page
- No automatic locale detection or redirection occurs
- Users must manually navigate to locale-prefixed routes (e.g., `/en/`, `/zh/about`)

## Screenshots

N/A (404 error pages are self-explanatory)

## Environment

- **Next.js Version**: [Check package.json]
- **Node Version**: [Your version]
- **OS**: All operating systems affected
- **Browser**: All browsers affected (Chrome, Firefox, Safari, Edge)
- **Deployment**: Affects both local development and production builds

### Related Changes
- **PR**: #448
- **Removed File**: `proxy.ts`

### Current Routing Structure
- All pages are located under `/app/[lang]/`
- Supported locales: `en` (English), `zh` (Chinese), `ja` (Japanese)
- Locale configuration: `/lib/i18n/config.ts`

### Technical Analysis
The `app/[lang]/page.tsx` file (lines 64-75) includes locale switching logic that redirects users to their saved locale preference. However, this code only executes when users are already on a locale-prefixed route. When accessing the root path `/`, this page component is never reached, resulting in a 404.

### Suggested Solutions
1. **Restore proxy.ts**: Revert the removal and keep the original locale detection logic
[...]

PR #462, opened 2025-12-30
fix: restore locale redirection using Next.js middleware

Biki-dev:
### Summary

Restored locale-based redirection by reintroducing the logic as a proper Next.js `middleware.ts`.

### Changes
- Added `middleware.ts` to handle locale detection and redirection
- Added cookie support to persist user locale

solves #458

### DIFF
--- a/proxy.ts
+++ b/proxy.ts
@@ -4,13 +4,7 @@ import type { NextRequest } from "next/server"
 import { NextResponse } from "next/server"
 import { i18n } from "./lib/i18n/config"
 
-function getLocale(request: NextRequest): string {
-    // Check for saved locale in cookie first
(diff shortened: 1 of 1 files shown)


broBinChen:
In Next.js version 16, the middleware file convention has been deprecated and should now be replaced with proxy.ts. For more details, please refer to: https://nextjs.org/docs/messages/middleware-to-proxy

DayuanJiang:
@broBinChen  Yes, I accidentally deleted the proxy.ts file. Putting it back should fix it.

Biki-dev:
@broBinChen Thanks, yes — that was my mistake. I used the old middleware name.

PR #463, opened 2025-12-30
feat: add copy button for tool call blocks

broBinChen:
Add a copy button in tool call headers to easily copy generated content.

Use case: Allow users to quickly copy generated XML/operations to use
as presets or for debugging purposes.

Changes:
- Add copy button in tool call header when expanded
- Support copying XML, JSON, and operations content

https://github.com/user-attachments/assets/d75b996a-1406-43ce-9b69-5556a4bdf278

### DIFF
--- a/components/chat-message-display.tsx
+++ b/components/chat-message-display.tsx
@@ -251,14 +251,12 @@ export function ChatMessageDisplay({
         Record<string, boolean>
     >({})
 
-    const copyMessageToClipboard = async (
+    const setCopyState = (
         messageId: string,
-        text: string,
-        isToolCall = false,
+        isToolCall: boolean,
+        isSuccess: boolean,
     ) => {
-        try {
-            await navigator.clipboard.writeText(text)
-
+        if (isSuccess) {
             if (isToolCall) {
                 setCopiedToolCallId(messageId)
                 setTimeout(() => setCopiedToolCallId(null), 2000)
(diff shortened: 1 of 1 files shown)


components/chat-message-display.tsx
```diff
@@ -245,12 +251,21 @@ export function ChatMessageDisplay({
...
 
-            setCopiedMessageId(messageId)
-            setTimeout(() => setCopiedMessageId(null), 2000)
+            if (isToolCall) {
+                setCopiedToolCallId(messageId)
+                setTimeout(() => setCopiedToolCallId(null), 2000)
+            } else {
+                setCopiedMessageId(messageId)
+                setTimeout(() => setCopiedMessageId(null), 2000)
+            }
```
DayuanJiang, line 266: These part of code seems appeared multiple times. Could be simplified with a helper or by abstracting the state updates.

DayuanJiang:
@broBinChen  Thanks for the contribution. I've added a minor comment about the code style—please check it. Otherwise, LGTM.

components/chat-message-display.tsx
broBinChen: Thanks for the review! I've refactored the code to extract the repeated copy state update logic into a `setCopyState` helper function.

broBinChen:
> @broBinChen Thanks for the contribution. I've added a minor comment about the code style—please check it. Otherwise, LGTM.

Thanks for the review! I've refactored the code to extract the repeated copy state update logic into a `setCopyState` helper function.

PR #465, opened 2025-12-30
 fix: persist chat message feedback state to localStorage

Biki-dev:
## 🎯 Summary

Fixes the issue where thumbs up/down feedback on chat messages is lost after page reload. Feedback state is now persisted to localStorage and properly restored on mount.

## 🔧 Changes

- Added `STORAGE_FEEDBACK_KEY` constant for localStorage persistence
- Initialize feedback state from localStorage on component mount

## 📝 Files Changed

- `components/chat-message-display.tsx` - Added localStorage sync for feedback state

### DIFF
--- a/components/chat-message-display.tsx
+++ b/components/chat-message-display.tsx
@@ -234,7 +234,27 @@ export function ChatMessageDisplay({
     const [copyFailedMessageId, setCopyFailedMessageId] = useState<
         string | null
     >(null)
-    const [feedback, setFeedback] = useState<Record<string, "good" | "bad">>({})
+    const STORAGE_FEEDBACK_KEY = "next-ai-draw-io-feedback"
+    const [feedback, setFeedback] = useState<Record<string, "good" | "bad">>(
+        () => {
+            if (typeof window !== "undefined") {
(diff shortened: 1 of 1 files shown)


DayuanJiang:
Thanks for the PR. 
The feedback is already being saved server-side when users click thumbs up/down (via /api/log-feedback), so the data isn't lost. This would only persist the visual highlight on the button, which isn't really necessary.

Biki-dev:
@DayuanJiang I see, but the user can’t see their like or dislike after reloading the page, which I’m trying to fix in this PR.
Should I proceed with a different approach, or is this fix not needed for now?

DayuanJiang:
It is actually not needed, since the app don't have history conversation feature, there is no need to perserver feedback state.

Biki-dev:
@DayuanJiang ohk , got it , now i close this pr then

DayuanJiang:
@Biki-dev  Thanks. Quick tip for next time: feel free to open an issue or message me first before starting work on a PR. That way we can discuss if/how the feature fits the project. Appreciate you contributing!

Biki-dev:
Yeah 👍

PR #467, opened 2025-12-30
feat(mcp-server): add DRAWIO_BASE_URL env for private deployments

liujin0506:
## Summary

Add support for configuring the diagrams.net embed URL in the MCP server via the `DRAWIO_BASE_URL` environment variable, allowing users to specify their own self-hosted draw.io instance for private deployments.

## Changes

- Add `DRAWIO_BASE_URL` environment variable support (defaults to `https://embed.diagrams.net`)
- Update iframe `src` to use the configured URL
- Update `postMessage` origin check to use the configured URL
- Add documentation for private deployment configuration in README

## Motivation

Organizations with strict security requirements need to use self-hosted draw.io instances. This change enables:

- **Data Privacy**: Sensitive diagrams stay within the organization's network
- **Air-gapped Environments**: Support for environments without external internet access
- **Compliance**: Meet regulatory requirements for data residency

## Usage

```json
{
  "mcpServers": {
    "drawio": {
      "command": "npx",
      "args": ["@next-ai-drawio/mcp-server@latest"],
      "env": { 
        "DRAWIO_BASE_URL": "https://drawio.your-company.com"
      }
    }
  }
}
```

## Testing

1. Start a local draw.io instance: `docker run -d -p 8080:8080 jgraph/drawio`
2. Set `DRAWIO_BASE_URL=http://localhost:8080`
3. Run the MCP server and verify diagrams load from the local instance

### DIFF
--- a/packages/mcp-server/src/http-server.ts
+++ b/packages/mcp-server/src/http-server.ts
@@ -17,6 +17,24 @@ import { log } from "./logger.js"
 const DRAWIO_BASE_URL =
     process.env.DRAWIO_BASE_URL || "https://embed.diagrams.net"
 
+// Extract origin (scheme + host + port) from URL for postMessage security check
+function getOrigin(url: string): string {
+    try {
+        const parsed = new URL(url)
+        return `${parsed.protocol}//${parsed.host}`
+    } catch {
+        return url // Fallback if parsing fails
+    }
+}
+
+const DRAWIO_ORIGIN = getOrigin(DRAWIO_BASE_URL)
+
+// Normalize URL for iframe src - ensure no double slashes
+function normalizeUrl(url: string): string {
+    // Remove trailing slash to avoid double slashes
+    return url.replace(/\/$/, '')
+}
+
 interface SessionState {
     xml: string
     version: number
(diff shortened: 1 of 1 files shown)


liujin0506:
Note: The Vercel check failure is unrelated to this PR - it appears the repository owner's Vercel account is blocked.

DayuanJiang:
Nice feature! One issue I noticed:

The origin check `e.origin !== '${DRAWIO_BASE_URL}'` will break if users add a trailing slash or path to their URL.

`e.origin` only returns scheme + host + port, not the path ([MDN](https://developer.mozilla.org/en-US/docs/Glossary/Origin)). So if someone sets `DRAWIO_BASE_URL=http://localhost:8080/` or `https://company.com/drawio`, the check will always fail.

liujin0506:
Good catch! Thank you for pointing this out. This issue has been fixed in commit `89a8cef`.

**Root Cause:**
The `e.origin` property from postMessage only returns `scheme + host + port`, without the path. So comparing it directly with `DRAWIO_BASE_URL` would fail when users configure URLs like:
- `http://localhost:8080/` (trailing slash)
- `https://company.com/drawio` (with path)

**Solution:**
1. Added `getOrigin()` function to extract the origin (scheme + host + port) from `DRAWIO_BASE_URL`
2. Created `DRAWIO_ORIGIN` constant for proper postMessage security check
3. Added `normalizeUrl()` to remove trailing slashes and avoid double slashes in iframe src

Now the origin check works correctly regardless of how users configure their `DRAWIO_BASE_URL`:
- `https://embed.diagrams.net/` → origin: `https://embed.diagrams.net` ✅
- `https://company.com/drawio` → origin: `https://company.com` ✅
- `https://drawio.example.com/custom/path/` → origin: `https://drawio.example.com` ✅

The fix is in the `feature/mcp-server-custom-drawio-url` branch.

DayuanJiang:
@liujin0506  Merged. Thanks and happy new year.

PR #470, opened 2025-12-31
style(chat-panel): Improve aiChat label display in collapsed panel

E66Crisp:
The previous implementation used `rotate(180deg)` which caused Chinese text to read from bottom to top. Removed/adjusted the rotation to follow natural CJK vertical reading order (top to bottom).

before:
image

after:
image


DayuanJiang:
Thanks, I didn't notice that.

PR #473, opened 2025-12-31
feat(i18n): sync Draw.io panel language with app locale

E66Crisp:
- Extract current language from URL pathname
- Pass lang parameter to DrawIoEmbed urlParameters
- Update component key to reload on language change


DayuanJiang:
Nice! Merged.

PR #490, opened 2026-01-01
feat: security and performance improvements

jasonkneen:
Security improvements:
- Add Zod schema validation for API request body (messages, xml, sessionId)
- Implement timing-safe access code comparison using crypto.timingSafeEqual to prevent timing attacks on access code validation
- Replace silent empty catch blocks with console.warn logging

Performance improvements:
- Fix memory leak: beforeunload listener now uses sessionIdRef to avoid re-registration on every sessionId change
- Add useMemo to flattenModels and findModelById in use-model-config to prevent unnecessary recalculations on every render

fix: address critical code review findings

- Add message history limit (MAX_PERSISTED_MESSAGES=100) to prevent localStorage overflow for active users
- Fix excessive re-renders during streaming by using chartXMLRef instead of chartXML in useEffect/useCallback dependencies
- Add security warning in model config dialog when API keys are stored in localStorage
- Fix postMessage wildcard origins in MCP server - now uses specific DRAWIO_ORIGIN instead of '*' to prevent cross-origin data leakage
- Improve TypeScript types: add MessageRole, ToolState, TextPart, FilePart, ToolPart interfaces; replace 6 any usages with proper types

### DIFF
--- a/app/[lang]/page.tsx
+++ b/app/[lang]/page.tsx
@@ -138,7 +138,12 @@ export default function Home() {
                 !isInitialRenderRef.current &&
                 newIsMobile !== isMobileRef.current
             ) {
-                saveDiagramToStorage().catch(() => {})
+                saveDiagramToStorage().catch((error) => {
+                    console.warn(
+                        "Failed to save diagram on viewport change:",
+                        error,
+                    )
+                })
                 setIsDrawioReady(false)
                 resetDrawioReady()
             }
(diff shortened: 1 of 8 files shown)


DayuanJiang:
Thanks for the contribution!
Could you please check the copilot's comment and resolve them first?
Also, there is a `bun.lock` in the commit, please remove it since we are using `npm` in this project.

PR #494, opened 2026-01-02
fix(mcp): prevent stuck spinner by initializing blank session state

yupeng0512:
### Problem
When opening the embedded draw.io preview for a brand new MCP session, the UI can show **Ready** but keep the draw.io iframe spinner running forever.

Root cause: the iframe is loaded with `spin=1`, but the embedded server returns `{ xml: null, version: 0 }` for new sessions, so the page never sends a `load(xml)` message and the spinner never stops.

### Fix
Initialize a minimal blank diagram in the embedded server for new `mcp-*` sessions (both on `/` and `/api/state`). This makes the first poll fetch a non-empty `xml` with `version > 0`, which triggers the existing `pendingXml` / `init` flow and stops the spinner without adding any client-side special cases.

### Why this approach
- Keeps the browser page logic unchanged and simple
- Makes session state consistent (no `xml: null` bootstrap edge case)
- Adds a conservative guard to avoid creating state for arbitrary session IDs

### Repro
1. Start MCP server
2. Open the browser URL for a fresh session
3. Before: Ready + infinite spinner
4. After: Ready + blank editable canvas (no stuck spinner)

### DIFF
--- a/packages/mcp-server/src/http-server.ts
+++ b/packages/mcp-server/src/http-server.ts
@@ -31,8 +31,7 @@ const DRAWIO_ORIGIN = getOrigin(DRAWIO_BASE_URL)
 
 // Minimal blank diagram used to bootstrap new sessions.
 // This avoids the draw.io embed spinner (spin=1) getting stuck when no `load(xml)` is ever sent.
-const DEFAULT_DIAGRAM_XML =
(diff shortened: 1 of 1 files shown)


DayuanJiang:
Thanks for the contribution. Neat fix. Merged.

PR #510, opened 2026-01-04
fix: save event is triggered frequently

coder-wangtao:
There is no need to remove the save button，you can do this to fix save dialog is triggered frequently

### DIFF
--- a/app/[lang]/page.tsx
+++ b/app/[lang]/page.tsx
@@ -17,8 +17,13 @@ const drawioBaseUrl =
     process.env.NEXT_PUBLIC_DRAWIO_BASE_URL || "https://embed.diagrams.net"
 
 export default function Home() {
-    const { drawioRef, handleDiagramExport, onDrawioLoad, resetDrawioReady } =
-        useDiagram()
+    const {
+        drawioRef,
+        handleDiagramExport,
+        onDrawioLoad,
+        resetDrawioReady,
+        setShowSaveDialog,
+    } = useDiagram()
     const router = useRouter()
     const pathname = usePathname()
     // Extract current language from pathname (e.g., "/zh/about" → "zh")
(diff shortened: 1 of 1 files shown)


DayuanJiang:
Thanks for the PR. I’m just wondering if we really need two Save buttons that do the same thing. What do you think?

coder-wangtao:
yes,I think you can delete the ChatPanel component save button,Because it is not easy to be found. the DrawIoEmbed component save button more user-friendly

DayuanJiang:
@coder-wangtao Yes, you are right, thanks of the idea.

coder-wangtao:
Glad to help you. If this helps you, you can merge and close this PR. thank you!

DayuanJiang:
@coder-wangtao 
Hey, after digging into the draw.io docs, I think I've got a much clearer picture of what's going on here.

Let's break down how saving works in draw.io:

1. draw.io Native "Save" Button (within the editor):
This one isn't actually for downloading files. Its job is to send the diagram's XML data back to our app (the "host") so we can save it wherever we need to. It won't trigger a file download.
2. This App's "Save" Button (Download icon in chat input):
Our app's "Save" button  is specifically for exporting the diagram as a file that the user can download.

The main problem we're hitting:
Any manual changes users make directly in the embedded draw.io editor aren't actually being saved or captured.

We can fix this by using draw.io's built-in autosave feature. This is actually how the official draw.io app handles things:

```
<DrawIoEmbed
    autosave={true}
    onAutoSave={(data) => {
        // This callback will automatically capture every edit!
    }}
/>
```

By turning autosave to true and using the onAutoSave callback, we'll automatically grab every change the user makes. This also means we won't need a separate "Save" button for this purpose anymore.

coder-wangtao:
@DayuanJiang I tried it. I didn't set autosave={true}, It can also save automatically.  App's "Save" Button (Download icon in chat input), I just feel it's very strange. People usually edit in the editor, and use draw.io Native "Save" Button to save. like me, They won't use  App's "Save" Button (Download icon in chat input)

Biki-dev:
> By turning autosave to true and using the onAutoSave callback
 
Yes, autosave is a good alternative solution. It helps users save changes locally.

PR #528, opened 2026-01-07
fix: add missing nanoid dependency

broBinChen:
image

add missing nanoid dependency for session storage


DayuanJiang:
@broBinChen  Thank you a lot for identifying this error. Merged.

PR #550, opened 2026-01-09
[Feature] Add setting for Enter/Ctrl+Enter to send messages

### LINKED ISSUE #549: [Feature] Add setting for Enter/Ctrl+Enter to send messages
## Feature Description
Add a user setting to configure the keyboard shortcut for sending messages - either "Enter to send" or "Ctrl+Enter to send".

## Problem Context
Different users have different preferences for message sending behavior:
- Some users prefer pressing Enter to quickly send messages (common in chat applications)
- Other users prefer Ctrl+Enter to avoid accidental sends when writing multi-line content (common in code editors and technical tools)

Currently there's no way to customize this behavior based on personal preference.

## Proposed Solution
Add a toggle in the Settings panel with two options:
- **Enter to send**: Press Enter to send, Shift+Enter for new line
- **Ctrl+Enter to send**: Press Enter for new line, Ctrl+Enter (or Cmd+Enter on Mac) to send

The setting should:
1. Be persisted in user preferences/localStorage
2. Have a sensible default (Enter to send is most common)
3. Be easily accessible in the Settings menu

## Alternatives Considered
- Always use Enter to send (current behavior in most chat apps)
- Always use Ctrl+Enter to send (safer for multi-line input)
- Auto-detect based on content length (complex and unpredictable)

## Additional Context
This is a common feature in messaging applications like:
- Slack (Settings > Send messages on Enter)
- Discord (similar setting available)
- Telegram (has this option)

maifeeulasad:
closes #549

### DIFF
--- a/components/chat-input.tsx
+++ b/components/chat-input.tsx
@@ -193,9 +193,7 @@ export function ChatInput({
     const [showHistory, setShowHistory] = useState(false)
     const [showUrlDialog, setShowUrlDialog] = useState(false)
     const [isExtractingUrl, setIsExtractingUrl] = useState(false)
-    const [sendShortcut] = useState(() =>
-        localStorage.getItem(STORAGE_KEYS.sendShortcut) || "ctrl-enter",
-    )
+    const [sendShortcut, setSendShortcut] = useState("ctrl-enter")
     // Allow retry when there's an error (even if status is still "streaming" or "submitted")
     const isDisabled =
         (status === "streaming" || status === "submitted") && !error
@@ -211,6 +209,24 @@ export function ChatInput({
     useEffect(() => {
         adjustTextareaHeight()
     }, [input, adjustTextareaHeight])
+
+    // Load send shortcut preference from localStorage and listen for changes
+    useEffect(() => {
+        const stored = localStorage.getItem(STORAGE_KEYS.sendShortcut)
+        if (stored) setSendShortcut(stored)
+
+        const handleChange = (e: CustomEvent<string>) =>
+            setSendShortcut(e.detail)
+        window.addEventListener(
+            "sendShortcutChange",
+            handleChange as EventListener,
+        )
+        return () =>
+            window.removeEventListener(
+                "sendShortcutChange",
+                handleChange as EventListener,
+            )
+    }, [])
 
     const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
         onChange(e)
--- a/components/settings-dialog.tsx
+++ b/components/settings-dialog.tsx
@@ -445,6 +445,11 @@ function SettingsContent({
                                     STORAGE_KEYS.sendShortcut,
                                     value,
                                 )
+                                window.dispatchEvent(
+                                    new CustomEvent("sendShortcutChange", {
+                                        detail: value,
+                                    }),
+                                )
                             }}
                         >
                             <SelectTrigger


maifeeulasad:
@DayuanJiang if you need some changes, please let me know.

components/chat-input.tsx
```diff
@@ -214,7 +215,14 @@ export function ChatInput({
     }
 
     const handleKeyDown = (e: React.KeyboardEvent) => {
-        if ((e.metaKey || e.ctrlKey) && e.key === "Enter") {
+        const sendShortcut =
+            localStorage.getItem(STORAGE_KEYS.sendShortcut) || "ctrl-enter"
+        const shouldSend =
+            sendShortcut === "enter"
+                ? e.key === "Enter" && !e.shiftKey
```
DayuanJiang: When in "Enter to send" mode, pressing Ctrl+Enter (or Cmd+Enter on Mac) should insert a newline, not send the message. Currently, the condition `e.key === "Enter" && !e.shiftKey` also matches Ctrl+Enter.

lib/i18n/dictionaries/en.json
```diff
@@ -104,6 +104,10 @@
         "closeProtectionDescription": "Show confirmation when leaving the page.",
         "diagramStyle": "Diagram Style",
         "diagramStyleDescription": "Toggle between minimal and styled diagram output.",
+        "sendShortcut": "Send Shortcut",
+        "sendShortcutDescription": "Choose how to send messages.",
+        "enterToSend": "Enter to send",
+        "ctrlEnterToSend": "Ctrl+Enter to send",
```
DayuanJiang: The label "Ctrl+Enter to send" doesn't account for Mac users who use Cmd+Enter. Consider using a platform-inclusive label like "Cmd/Ctrl+Enter to send".

components/chat-input.tsx
maifeeulasad: Just resolved!

maifeeulasad:
@DayuanJiang care to take another look please?

components/chat-input.tsx
```diff
@@ -214,7 +215,14 @@ export function ChatInput({
     }
 
     const handleKeyDown = (e: React.KeyboardEvent) => {
-        if ((e.metaKey || e.ctrlKey) && e.key === "Enter") {
+        const sendShortcut =
+            localStorage.getItem(STORAGE_KEYS.sendShortcut) || "ctrl-enter"
```
DayuanJiang: It seems `localStorage.getItem` is called on every keystroke in `handleKeyDown`. Consider reading it once on mount using a `useRef` + `useEffect` instead.

maifeeulasad: Hey man, just thinking out loud.

If we take the `useMemo` route, we will be reading the state only once, on component mount. Something like this.
```
const sendShortcut = useMemo(() => {
  return localStorage.getItem(STORAGE_KEYS.sendShortcut) || "ctrl-enter";
}, []);
```
If this changes value on the background we won't be notified. It's the same if we use `useRef`.

Then we have another route. Something like:
```
const [sendShortcut, setSendShortcut] = useState(
  () => localStorage.getItem(STORAGE_KEYS.sendShortcut) || "ctrl-enter"
);
```

And then we can monitor it with:
```
useEffect(() => {
  const handler = () => {
    setSendShortcut(
      localStorage.getItem(STORAGE_KEYS.sendShortcut) || "ctrl-enter"
    );
  };

  window.addEventListener("storage", handler);
  return () => window.removeEventListener("storage", handler);
}, []);
```

Personally I think, this would be a great solution, although over-engineering. But correct one.

@DayuanJiang  What do you think?? Pick one, I will happily push that solution.

DayuanJiang: Thansk for the proposal, the first option seems simple and good, but why not just use `useState` since it is a state?

maifeeulasad: Cool! Will do that

DayuanJiang:
@maifeeulasad Thanks for the fix. I tested it locally and found it is better to make it work without refreshing the page to keep the setting behavior consistent. I fixed it myself. Merged.

maifeeulasad:
@DayuanJiang really appreciate it. I was nice contributing to this open source project and working with this really collaborative team! 🥂

PR #456, opened 2025-12-30
feat: add EdgeOne Pages as AI provider

zhoujie0531:
## Summary
Add EdgeOne Pages as a new AI provider option, enabling users to use EdgeOne's Edge AI service with DeepSeek models **without requiring an API key**.

## Changes
- Add EdgeOne provider option in model selector and configuration
- Implement OpenAI-compatible Edge Function (`/api/edgeai/chat/completions`) for EdgeOne Pages
- Support DeepSeek models: `deepseek-v3.2`
- Add model validation support for EdgeOne provider

## Features
- **No API key required** - uses EdgeOne's built-in AI service
- OpenAI-compatible API interface
- Streaming response support with native tool calling
- Automatic deployment to EdgeOne Pages

## Files
- `edge-functions/api/edgeai/chat/completions.ts` - Edge Function implementation
- `edgeone.json` - EdgeOne Pages configuration
- `lib/ai-providers.ts` - Provider registration
- `components/model-selector.tsx` - UI integration
- `app/api/validate-model/route.ts` - Model validation

PR #574, opened 2026-01-11
Feat/gcp vertex ai

### LINKED ISSUE #572: [Feature] Add GCP Vertex AI as Provider
## Feature Description

Add Google Cloud Platform (GCP) Vertex AI as a new AI provider option, allowing users to use Gemini models via their GCP project.

## Problem Context

Currently, users can use Google's Gemini models via the public Google Generative AI API (`@ai-sdk/google`). However, enterprise users often need to use Vertex AI for:
- Data residency requirements
- VPC Service Controls
- Integration with existing GCP infrastructure
- Usage of private endpoints

## Proposed Solution

Add Vertex AI as a new provider using `@ai-sdk/google-vertex` package from the Vercel AI SDK.

### Environment Variables

```
GOOGLE_VERTEX_PROJECT=your-gcp-project-id
GOOGLE_VERTEX_LOCATION=us-central1  # optional, defaults to us-central1
```

### Reference

- AI SDK Google Vertex Provider: https://sdk.vercel.ai/providers/ai-sdk-providers/google-vertex
- Similar PR for adding ModelScope: #521
- Similar PR for adding EdgeOne: #456

## Additional Context

The implementation pattern is similar to existing providers in `lib/ai-providers.ts`.

ElshadHu:
## What I did

I added Google Vertex AI as a new AI provider.

### 1. Dependency

I installed `@ai-sdk/google-vertex` from the Vercel AI SDK.

### 2. Backend

I added Vertex AI provider with `createVertex()`. It uses `GOOGLE_VERTEX_PROJECT` and `GOOGLE_VERTEX_LOCATION` env vars. No API key needed - it uses GCP service account auth.

### 3. Route

I added validation for Vertex AI models. It checks if the project env var is set, then validates the model.

### 4. UI (like EdgeOne)

I made Vertex AI work like EdgeOne - no API key field, just a Test button. Users add models and click Test to validate.

### 5. Testing

I tested it with my GCP project. I added `gemini-2.5-pro` model and it worked. See screenshots below.

## How to use

```bash
AI_PROVIDER=vertexai
AI_MODEL=gemini-2.5-pro
GOOGLE_VERTEX_PROJECT=your-project-id
```

Closes #572 

### Screenshot
image

### DIFF
--- a/app/api/validate-model/route.ts
+++ b/app/api/validate-model/route.ts
@@ -176,6 +176,7 @@ export async function POST(req: Request) {
             case "vertexai": {
                 const vertex = createVertex({
                     apiKey: vertexApiKey,
+                    ...(baseUrl && { baseURL: baseUrl }),
                 })
                 model = vertex(modelId)
                 break
--- a/lib/ai-providers.ts
+++ b/lib/ai-providers.ts
@@ -293,19 +293,21 @@ function buildProviderOptions(
             break
         }
         case "vertexai": {
-            // Google Vertex supports thinking config for thinking-enabled models
-            // Only models with "thinking" in their name support this feature
-            const isThinkingModel = modelId?.toLowerCase().includes("thinking")
+            const thinkingBudget = parseIntSafe(
+                process.env.GOOGLE_VERTEX_THINKING_BUDGET,
+                "GOOGLE_VERTEX_THINKING_BUDGET",
+                1024,
+                100000,
+            )
+            const thinkingLevel = process.env.GOOGLE_VERTEX_THINKING_LEVEL
 
-            if (isThinkingModel) {
-                const thinkingBudget = parseIntSafe(
-                    process.env.GOOGLE_VERTEX_THINKING_BUDGET,
-                    "GOOGLE_VERTEX_THINKING_BUDGET",
-                    1024,
-                    100000,
-                )
-                const thinkingLevel = process.env.GOOGLE_VERTEX_THINKING_LEVEL
-
+            if (
+                modelId &&
+                (modelId.includes("gemini-2") ||
+                    modelId.includes("gemini-3") ||
+                    modelId.includes("gemini2") ||
+                    modelId.includes("gemini3"))
+            ) {
                 const thinkingConfig: Record<string, any> = {
                     includeThoughts: true,
                 }
@@ -313,14 +315,17 @@ function buildProviderOptions(
                 const isGemini3 =
                     modelId?.includes("gemini-3") ||
                     modelId?.includes("gemini3")
+                const isGemini25 =
+                    modelId?.includes("2.5") || modelId?.includes("2-5")
+
                 if (isGemini3 && thinkingLevel) {
-                    // Gemini 3: Use thinkingLevel (minimal/low/medium/high)
+                    // Vertex AI provider in AI SDK supports more granular levels (minimal/low/medium/high)
                     thinkingConfig.thinkingLevel = thinkingLevel as
                         | "minimal"
                         | "low"
                         | "medium"
                         | "high"
-                } else if (!isGemini3 && thinkingBudget) {
+                } else if (isGemini25 && thinkingBudget) {
                     thinkingConfig.thinkingBudget = thinkingBudget
                 }
                 options.google = { thinkingConfig }
@@ -532,6 +537,7 @@ export function getAIModel(overrides?: ClientOverrides): ModelConfig {
     if (
         overrides?.baseUrl &&
         !overrides?.apiKey &&
+        !(overrides?.provider === "vertexai" && overrides?.vertexApiKey) &&
         overrides?.provider !== "edgeone"
     ) {
         throw new Error(


Biki-dev:
i will check it soon , thanks.

ElshadHu:
> The implementation is clean, but while reviewing it I found an issue that could be improved. The client-side model configuration is not implemented correctly. Don’t you think so?
> 
> If a user tries to configure the model on the client side, they can only change the display name. However, if they need to change `GOOGLE_VERTEX_PROJECT=your-gcp-project-id` or `GOOGLE_VERTEX_LOCATION=us-central1`, they must configure these on the server side via environment variables.
> 
> Please check this.

I got it. Vertex ai currently relies on server-side environment variables, which does not allow users from using their own GCP projects. I'm updating the client side to allow users to optionally specify their own Project ID and Location.

ElshadHu:
Hey @Biki-dev , I updated the provider interfaces to include `vertexProject` and `vertexLocation` as optional fields. I chose this approach because this allows to pass these values optionally, keeping the configuration pattern consistent across different providers. If there is any issue, please, let me know.

Biki-dev:
### Issue: Client-side Vertex AI Configuration Not Applied During Chat Requests

First of all, this PR looks good and is ready to merge 👍

The overall implementation is solid.

However, while testing, I noticed a related issue that might be worth addressing (if feasible):

When `GOOGLE_VERTEX_PROJECT` is configured **only via the client-side Settings UI** (and not in the `.env` file), chat generation fails with the following error:

```json
{
  "error": "Project ID is required for Vertex AI. Please configure it in Settings or set GOOGLE_VERTEX_PROJECT environment variable."
}
```

### Observed Behavior
- `vertexProject` and `vertexLocation` configured in the UI are validated correctly.
- During actual chat generation, these values are **not sent** from the client to the server .
- As a result, the server behaves as if the Vertex configuration is missing.

This suggests the UI-based configuration is not fully wired into the chat request flow.

@DayuanJiang I think the current implementation is still great and safe to merge as-is. As PR #456 has a similar base-url issue, that model also needs a client-side fix in a different PR.

That would be awesome if this issue solved—but no blocker from my side.

Great work overall 👏

DayuanJiang (commented):
@ElshadHu @Biki-dev  Thanks for the implementation and review!

I have found some issues:
1. The config should not be using the local json file since most of the time we will deploy it in the edge runtime like Vercel, which cannot access the file system. The best way is to use API key, please check https://ai-sdk.dev/providers/ai-sdk-providers/google-vertex#express-mode
2. The user side model config panel should also use api key as input. Also, it needs a test button to test the api (currently no button)
3. For thinking settings: Gemini 3 uses thinkingLevel (minimal/low/medium/high). Please check https://cloud.google.com/vertex-ai/generative-ai/docs/thinking

ElshadHu:
> @ElshadHu @Biki-dev Thanks for the implementation and review!
> 
> I have found some issues:
> 
> 1. The config should not be using the local json file since most of the time we will deploy it in the edge runtime like Vercel, which cannot access the file system. The best way is to use API key, please check https://ai-sdk.dev/providers/ai-sdk-providers/google-vertex#express-mode
> 2. The user side model config panel should also use api key as input. Also, it needs a test button to test the api (currently no button)
> 3. For thinking settings: Gemini 3 uses thinkingLevel (minimal/low/medium/high). Please check https://cloud.google.com/vertex-ai/generative-ai/docs/thinking

İ will check them out.

ElshadHu:
Hey @Biki-dev @DayuanJiang  

I've implemented the Express Mode changes (API key input, test button) and updated the thinking settings. However, I'm stuck on the authentication.

**The problem:**

`@ai-sdk/google-vertex` with `apiKey` option routes to `aiplatform.googleapis.com`, which rejects API keys with:
```
Permission 'aiplatform.endpoints.predict' denied
```

I tested with a valid Express Mode key and confirmed:
- `aiplatform.googleapis.com` -> 1/3 times error happens
- `generativelanguage.googleapis.com` -> no error

I've read the docs but Vertex AI auth with Express Mode is tricky. Can you help me figure out the correct approach here?

Biki-dev:
Hey @ElshadHu,
The authentication error you're seeing happens probably because the current code is trying to use the standard Google Cloud Service Account flow instead of the Express Mode flow. Even if you have an API key, the SDK won't use it correctly unless it is explicitly passed into the provider's initialization.
To fix this, you need to ensure that the "apiKey" is being passed directly into the "createVertex" function in both your main AI provider logic and your validation route. When the Vercel AI SDK receives an API key in the configuration, it automatically switches its internal logic to use the Express Mode endpoint and authentication method. This will resolve the permission denied error because Express Mode is specifically designed to authorize requests using these keys rather than GCP IAM permissions.
Once the API key is properly passed from the UI to the backend initialization, the "aiplatform.endpoints.predict" error should disappear.
let me know your thought

ElshadHu:
> Hey @ElshadHu, The authentication error you're seeing happens probably because the current code is trying to use the standard Google Cloud Service Account flow instead of the Express Mode flow. Even if you have an API key, the SDK won't use it correctly unless it is explicitly passed into the provider's initialization. To fix this, you need to ensure that the "apiKey" is being passed directly into the "createVertex" function in both your main AI provider logic and your validation route. When the Vercel AI SDK receives an API key in the configuration, it automatically switches its internal logic to use the Express Mode endpoint and authentication method. This will resolve the permission denied error because Express Mode is specifically designed to authorize requests using these keys rather than GCP IAM permissions. Once the API key is properly passed from the UI to the backend initialization, the "aiplatform.endpoints.predict" error should disappear. let me know your thought

Thanks for the detailed explanation! That makes sense, I'll update the code to pass the apiKey directly into the `createVertex` function. I'll let you know once I've tested it.

ElshadHu:
Hey @Biki-dev @DayuanJiang , is there any updates on that PR ?

DayuanJiang:
@ElshadHu Could you please address all Copilot comments? If a comment is correct, please fix it; if not, simply click the `resolve conversation` button.

ElshadHu:
I fixed the issues after copilot suggestions. I think it is done. @DayuanJiang

DayuanJiang:
Hi @Biki-dev  this pr is good to merge. Could you change the review state?

Biki-dev:
@DayuanJiang good to go

Issue #585, opened 2026-01-15
[Enhancement] Set focus to input area after clicking 'Start Fresh Chat'


DayuanJiang:
## Current Behavior
When clicking the "Start Fresh Chat" button, the focus shifts to the draw.io panel instead of staying in the chat area.
![Image](https://github.com/user-attachments/assets/2539186b-3d84-46b4-86a1-ee9e99deffcf)

## Proposed Enhancement
After clicking "Start Fresh Chat", the focus should automatically be set to the chat input area, allowing users to immediately start typing their new message.

## Motivation
This improves user experience by:
- Reducing the number of clicks/actions needed to start a new conversation
- Providing a more intuitive flow where users can immediately begin typing after starting a fresh chat
- Keeping the user's attention in the chat interface where they intend to interact

## Additional Context
This likely requires calling `.focus()` on the input element after the fresh chat action is completed.

PR #586, opened 2026-01-15
Set focus to input area after clicking Start Fresh Chat

Biki-dev:
•  Add ref-based focus control to ChatInput so it can be programmatically focused from ChatPanel.
•  When starting a “New Chat”, clear existing messages, diagram state, draft input, and URL, then auto-focus the chat input.
•  Ensure half-typed text is discarded when starting a fresh chat, so users always begin with a clean input.

solves #585

### DIFF
--- a/components/chat-input.tsx
+++ b/components/chat-input.tsx
@@ -9,14 +9,7 @@ import {
     Send,
 } from "lucide-react"
 import type React from "react"
-import {
-    forwardRef,
-    useCallback,
-    useEffect,
-    useImperativeHandle,
-    useRef,
-    useState,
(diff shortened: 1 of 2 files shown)


DayuanJiang (commented):
@Biki-dev  It seems you are using  the older version of the main, could you rebase to the latest main branch?

Biki-dev:
I rebased onto the latest main to resolve conflicts, and now the PR diff includes additional changes (likely due to upstream updates).
Can you confirm if this looks acceptable to merge, or should I create a new clean PR to keep the diff minimal? Thank you!

PR #591, opened 2026-01-16
Improve UX: Disable save button when diagram is empty

broBinChen:
This PR improves user experience by disabling the save button when the diagram canvas is empty. Currently, users can save an empty diagram, which results in downloading an empty PNG or a meaningless SVG file with no actual content.


DayuanJiang:
Neat fix, thanks.

PR #592, opened 2026-01-16
fix: add missing @opentelemetry/api dependency

broBinChen:
image

add missing @opentelemetry/api dependency for langfuse.ts


DayuanJiang:
@broBinChen  Thanks a lot!

Issue #594, opened 2026-01-16
[Enhancement] Add i18n support for Electron menu


DayuanJiang:
## Current Behavior
The Electron app menu (File, Edit, View, Configuration, Window, Help) is hardcoded in English in `electron/main/app-menu.ts`, even when the rest of the UI is displayed in another language (Chinese/Japanese).

![Menu in English while UI is in Chinese](https://github.com/user-attachments/assets/placeholder)

## Proposed Enhancement
Add internationalization (i18n) support for the Electron menu to match the UI language. The menu should support:
- English
- Chinese (中文)
- Japanese (日本語)

## Motivation
- Consistent user experience across the entire application
- Better accessibility for non-English speakers
- The web UI already supports multiple languages, but the Electron menu does not

## Implementation Notes
The menu is defined in `electron/main/app-menu.ts`. Key changes needed:

1. **Detect system locale** - Use `app.getLocale()` to get the system language
2. **Add translation maps** - Create translation objects for each supported language
3. **Replace hardcoded strings** - Update labels like "File", "Edit", "View", "Configuration", "Window", "Help", "Settings", "Documentation", "Report Issue", etc.

Note: Menu items using Electron's `role` property (like `role: "undo"`, `role: "copy"`) already use OS language automatically.

## Additional Context
Related files:
- `electron/main/app-menu.ts` - Main menu configuration
- Existing UI translations in the web app could be referenced for consistency

PR #596, opened 2026-01-16
[Enhancement] Add i18n support for Electron menu

Biki-dev:
## Problem
The Electron app menu was hardcoded in English, creating inconsistency when the rest of the UI displayed in other languages.

## Solution
Added internationalization (i18n) support to the Electron menu, matching the UI language with support for:
- English
- Chinese (简体中文)
- Japanese (日本語)

### 1. Created `menu-i18n.ts`
- Centralized translation module with type-safe interface
- Automatic locale detection from system language (`app.getLocale()`)
- Graceful fallback to English for unsupported locales

### 2. Updated `app-menu.ts`
- Replaced hardcoded English strings with translations
- Preserved Electron's built-in role translations (Undo, Copy, Paste etc.)
- Maintained all existing functionality and keyboard shortcuts

## Testing
The menu now automatically displays in the appropriate language based on system settings. No user configuration required.

#594

### DIFF
--- a/electron/main/menu-i18n.ts
+++ b/electron/main/menu-i18n.ts
@@ -10,15 +10,9 @@ export type MenuLocale = "en" | "zh" | "ja"
 export interface MenuTranslations {
     // App menu (macOS only)
     settings: string
-    services: string
-    hide: string
-    hideOthers: string
-    unhide: string
-    quit: string
 
     // File menu
     file: string
-    close: string
 
     // Edit menu
     edit: string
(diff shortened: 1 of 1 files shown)


electron/main/menu-i18n.ts
```diff
@@ -0,0 +1,178 @@
...
+    if (normalized === "zh") return "zh"
+    if (normalized === "ja") return "ja"
+    return "en"
+}
+
+/**
+ * Get locale from stored preference or system default
+ * Checks localStorage for user's language preference first
+ */
+export function getPreferredLocale(appLocale: string): MenuLocale {
```
DayuanJiang, line 153: Defined and never used function

Biki-dev:
### Sync Electron Menu Language with Web App Preference

This PR ensures the Electron **menu language stays consistent** with the language selected inside the web app (instead of always using the system locale).

### ✅ Key Changes
- Persist user's language preference in the **Electron config file** (alongside presets)
- Implement **`getPreferredLocale()`** to load the saved locale first, with fallback to `app.getLocale()`
- Add **IPC handlers** to sync language updates from the web app → Electron main process
- **Rebuild the menu immediately** when the user changes language preference

### 🎯 Result
Electron menu and app UI now always display in the **same user-selected language**, eliminating mismatch caused by system locale differences.

PR #598, opened 2026-01-16
feat: Update some tool demos in english about page

khanhthanhdev:
@DayuanJiang  
I add some examples in about page. It works well in desktop and mobile view
image
image


Biki-dev (commented):
Hello @khanhthanhdev,
I checked your PR and noticed that the changes were made only for the English About page. However, the About pages for the other two languages also need the same update.
It would be great if you could apply the same changes to all three files.
Also, GitHub Copilot suggested a few changes—please review them. If the suggestions are correct, implement them; otherwise, feel free to ignore them.

Thank you!

DayuanJiang:
@khanhthanhdev  Thanks for the contribution, I also add the example to readme.md

PR #617, opened 2026-01-20
fix: replace ring with border for provider selection state

broBinChen:
Fixes incomplete border rendering when a provider is selected in the model configuration dialog.

before: 
image

after:
image

### DIFF
--- a/components/settings-dialog.tsx
+++ b/components/settings-dialog.tsx
@@ -432,7 +432,7 @@ function SettingsContent({
                         >
                             <SelectTrigger
                                 id="send-shortcut-select"
-                                className="w-[170px] h-9 rounded-xl"
+                                className="w-auto h-9 rounded-xl"
                             >
                                 <SelectValue />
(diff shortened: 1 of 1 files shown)


Biki-dev:
image

Can you fix this as well? The full line wasn’t visible — it just needs a width adjustment. Please fix it. It’s a small issue, so I don’t want to open a new issue for this.

broBinChen:
> image
> Can you fix this as well? The full line wasn’t visible — it just needs a width adjustment. Please fix it. It’s a small issue, so I don’t want to open a new issue for this.

Thanks for catching that! I've fixed the width issue for the send shortcut selector. I changed it from a fixed width to `w-auto` so it can automatically adjust based on the content length, which should work better across different languages as well. The full text should now be visible without being cut off.
image

DayuanJiang:
@broBinChen  @Biki-dev  Thanks a lot!

PR #626, opened 2026-01-22
feat(electron): bundle draw.io for offline support

DayuanJiang:
## Problem

The Electron app requires internet access because draw.io loads from `https://embed.diagrams.net`. In offline environments, the app fails to work.

## Solution

Bundle draw.io static files (~140MB) with the Electron app and serve them locally.

## Changes

- Download draw.io v29.3.5 static files during CI build
- Detect Electron and use local draw.io files instead of CDN
- Add `offline=1` parameter to disable external service calls in draw.io
- Skip `/drawio` path from i18n middleware redirect
- Add `public/drawio/` to .gitignore (files downloaded during build, not committed)

Tested offline and confirmed working.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "next-ai-draw-io",
-    "version": "0.4.12-beta.1",
+    "version": "0.4.12-beta.2",
     "lockfileVersion": 3,
     "requires": true,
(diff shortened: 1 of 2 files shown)


DayuanJiang:
Closing to create a clean PR based on main

PR #641, opened 2026-01-24
feat(prompt): add language-aware response rules with english fallback

### LINKED ISSUE #635: Add language matching instruction to system prompt for non-English users
### Problem
When using certain AI models (particularly Gemini), the assistant responds in English even when the user writes in a different language (e.g., Chinese, Japanese). This creates a poor user experience for non-English speakers.

**Fixes #630**

### Root Cause
The system prompt in `lib/system-prompts.ts` does not include any instruction telling the AI to respond in the user's language. While some models (like Claude) automatically match the user's language, others (like Gemini) default to English without explicit instruction.

### Proposed Solution
Add a language matching instruction to the `DEFAULT_SYSTEM_PROMPT` in `lib/system-prompts.ts`.

**Location:** `lib/system-prompts.ts` (line ~11-15)

**Suggested change:** Add this instruction after the first paragraph:

```
Respond in the same language the user uses. Match their language naturally without mentioning this instruction.
```

### Files to Modify
- `lib/system-prompts.ts` - Add language instruction to `DEFAULT_SYSTEM_PROMPT`

### Acceptance Criteria
- [ ] Language instruction added to the system prompt
- [ ] Test with a non-English prompt (e.g., Chinese) to verify the model responds in the same language
- [ ] No changes to existing diagram generation functionality

### Getting Started
1. Fork and clone the repository
2. Run `npm install` to install dependencies
3. Edit `lib/system-prompts.ts`
4. Test locally with `npm run dev`
5. Submit a PR

### Notes for Contributors
- Keep the instruction brief (saves tokens)
- Place it early in the prompt for higher priority
- The instruction should work across all supported models (OpenAI, Anthropic, Google, etc.)

Subhajeetch:
### Related Issue

Closes #635

## Summary

This PR updates the system prompt to instruct the model to always respond in the same language as the user's last message, improving usability for non-English users.

## New Behavior

- Responses follow the user's language automatically
- Language switches are handled mid-conversation
- If the model cannot confidently respond in the user's language, it falls back to English with a short apology
- Diagram tool usage and formatting rules remain unchanged

## Why
Currently, the assistant defaults to English even when users communicate in other languages. Adding explicit language-matching rules at the system-prompt level ensures more inclusive and predictable behavior across multilingual users.

## Scope of Change
- System prompt only (no runtime or API behavior changes)
- No impact on diagram generation or tool invocation logic

### DIFF
--- a/lib/system-prompts.ts
+++ b/lib/system-prompts.ts
@@ -9,17 +9,9 @@
 // Default system prompt (~1900 tokens) - works with all models
 export const DEFAULT_SYSTEM_PROMPT = `
 You are an expert diagram creation assistant specializing in draw.io XML generation.
-
-## Language Handling Rules
-- ALWAYS respond in the same language as the user's last message.
-- If the user switches languages, immediately switch to that language.
-- If the model cannot confidently respond in the user's language, reply in English and briefly say:
-  "Sorry, I can't fully support that language yet, so replying in English."
-- Tool calls (display_diagram / edit_diagram) must still follow all rules, but any explanatory text before tool calls must match the user's language.
-- If the user's message mixes multiple languages, prefer the dominant language used for instructions.
-
 Your primary function is chat with user and crafting clear, well-organized visual diagrams through precise XML specifications.
 You can see images that users upload, and you can read the text content extracted from PDF documents they upload.
+ALWAYS respond in the same language as the user's last message.
 
 When you are asked to create a diagram, briefly describe your plan about the layout and structure to avoid object overlapping or edge cross the objects. (2-3 sentences max), then use display_diagram tool to generate the XML.
 After generating or editing a diagram, you don't need to say anything. The user can see the diagram - no need to describe it.


lib/system-prompts.ts
```diff
@@ -9,6 +9,15 @@
...
 export const DEFAULT_SYSTEM_PROMPT = `
 You are an expert diagram creation assistant specializing in draw.io XML generation.
+
+## Language Handling Rules
+- ALWAYS respond in the same language as the user's last message.
+- If the user switches languages, immediately switch to that language.
+- If the model cannot confidently respond in the user's language, reply in English and briefly say:
+  "Sorry, I can't fully support that language yet, so replying in English."
+- Tool calls (display_diagram / edit_diagram) must still follow all rules, but any explanatory text before tool calls must match the user's language.
+- If the user's message mixes multiple languages, prefer the dominant language used for instructions.
```
DayuanJiang: Thanks for the contribution!

I'd suggest keeping only line 14, as this is a straightforward task. More detailed instructions might distract the LLM from its main objective—generating the diagram.

Subhajeetch: Thanks for the feedback!

I've simplified the prompt to keep only the single language-matching instruction as suggested, so it stays focused on diagram generation.
Thanks again :)

PR #659, opened 2026-01-29
fix(i18n): add missing translation keys in zh-Hant.json

broBinChen:
## Summary
Fix missing translation keys in Traditional Chinese (zh-Hant) locale file.

## Problem
The `zh-Hant.json` dictionary was missing two required translation keys in the `settings` section:
- `apiKeysModels`
- `apiKeysModelsDescription`

This caused a TypeScript type compatibility error (ts(2719)) because the type definition expects all locale files to have the same structure.
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "next-ai-draw-io",
-    "version": "0.4.12-beta.5",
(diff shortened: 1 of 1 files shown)


DayuanJiang:
@broBinChen  Thanks for the contribution.

Issue #637, opened 2026-01-24
[Enhancement] Pin 'Configure Models...' button to bottom of model selector dropdown


DayuanJiang:
## Current Behavior
The "Configure Models..." option is placed at the bottom of the model selector dropdown list, but it scrolls with the rest of the content. When there are many models, users may not notice or find where to configure new models.

## Proposed Enhancement
Fix/pin the "Configure Models..." button to the bottom of the dropdown, so it remains visible regardless of scrolling position.

## Motivation
- **Improved discoverability**: New users can easily find where to configure custom models
- **Better UX**: Users don't need to scroll to the bottom to access configuration
- **Consistent behavior**: Similar to how many apps pin important actions at fixed positions

## Screenshots / Mockups
The red annotation shows the desired behavior - the "Configure Models..." option should be fixed to the bottom of the dropdown.

<img width="1038" height="720" alt="Image" src="https://github.com/user-attachments/assets/2000a4a1-550d-45ef-820c-e47c8e13e925" />

## Additional Context
This is a UI/UX enhancement that would improve the overall user experience for model configuration.

Gideon1107:
I would like to work on this . Please assign it to me

Biki-dev:
@Gideon1107 thanks

PR #665, opened 2026-01-30
enhancement: Pin Configure Models button to bottom of model selector …

Gideon1107:
…(#637)

Keep "Configure Models..." and info text fixed at bottom of dropdown when the model list scrolls. Wire button to open Model Config dialog.

### DIFF
--- a/components/model-selector.tsx
+++ b/components/model-selector.tsx
@@ -440,7 +440,7 @@ export function ModelSelector({
                                     </ModelSelectorItem>
                                 </div>
                             )}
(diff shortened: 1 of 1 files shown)


Gideon1107:
Addressed in the latest commit: switched to ModelSelectorItem for keyboard nav, added z-10 on the footer so it sits above the shadow, and moved spacing to the wrapper with px-3 py-2. Thanks for the review.

DayuanJiang:
@Gideon1107  Thanks for the contribution and fix, I find there was a minor issue about the space and fixed it. Now it can be merged.

PR #694, opened 2026-02-10
Fix: return clear error for PDF URLs in content extraction

### LINKED ISSUE #693: [Bug]: PDF URL extraction returns 400 error
## Description

When using the "Extract Content from URL" feature with PDF files, the application returns a 400 error. The backend console shows a deprecation warning about `url.parse()` at the same time.

## Steps to Reproduce

1. Click "Extract content from URL" button in the chat input
2. Paste a PDF URL (e.g., `https://www.nerc.com/globalassets/our-work/reports/special-reports/nerc_cip_roadmap_01122026.pdf`)
3. Click "Extract"
4. Observe 400 error in browser network tab
5. Check backend console for deprecation warning

## Expected Behavior

The application should either:
- Successfully extract text content from the PDF URL, OR
- Return a clear, user-friendly error message indicating that PDF URLs are not supported and the user should upload the file directly instead

## Actual Behavior

- Returns a generic 400 error to the user
- Backend console shows: `(node:1) [DEP0169] DeprecationWarning: url.parse() behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead.`
- The backend appears to successfully fetch the PDF (TLS connection established, data transferred), but fails during the processing/extraction step

## Root Cause Analysis

Based on code review of `app/api/parse-url/route.ts`:

1. The API uses `@extractus/article-extractor` which is designed for HTML/text content, not PDFs
2. When the URL returns `application/pdf` content type, the extractor fails to parse it
3. The current implementation doesn't check Content-Type before attempting extraction
4. This results in a generic 400 error: "Could not extract content from URL"

## Proposed Solution

Add Content-Type validation in `/app/api/parse-url/route.ts` before passing to the extractor:

```typescript
// After fetching the URL, check Content-Type header
const headResponse = await fetch(url, {
    method: 'HEAD',
    headers: {
        "User-Agent": "Mozilla/5.0 (compatible; NextAIDrawio/1.0)",
    },
    signal: controller.signal,
});

const contentType = headResponse.headers.get('content-type');
if (contentType?.includes('application/pdf')) {
    return NextResponse.json(
        { error: "PDF URLs are not supported. Please download and upload the PDF file directly." },
        { status: 400 },
    );
}
```

## Additional Notes

- PDF file uploads work correctly - this issue only affects PDF URLs
- The extractor successfully handles HTML/text content URLs

## Related discusstion

- #690 

## Related pr
- #514

ElshadHu:
### What I Changed

I went with 422 instead of 400 here  because I thoguh the request itself is valid , but the content at the URL can't be processed. 422 fits that better than  400. Happy to change it if you prefer something else though.

I used a separate `AbortController` with a 3s timeout so it doesn't interfere with the main extraction timeout. If the HEAD request fails for any reason (server doesn't support HEAD, timeout, network error), it logs a warning and falls through to the extractor 

## Screenshot
image

Closes #693

### DIFF
--- a/app/api/parse-url/route.ts
+++ b/app/api/parse-url/route.ts
@@ -5,6 +5,7 @@ import { allowPrivateUrls, isPrivateUrl } from "@/lib/ssrf-protection"
 
 const MAX_CONTENT_LENGTH = 150000 // Match PDF limit
 const EXTRACT_TIMEOUT_MS = 15000
+const USER_AGENT = "Mozilla/5.0 (compatible; NextAIDrawio/1.0)"
 
 export async function POST(req: Request) {
     try {
@@ -34,18 +35,14 @@ export async function POST(req: Request) {
                 { status: 400 },
             )
         }
+        const headController = new AbortController()
+        const headTimeout = setTimeout(() => headController.abort(), 3000)
         try {
-            const headController = new AbortController()
-            const headTimeout = setTimeout(() => headController.abort(), 3000)
-
             const headResponse = await fetch(url, {
                 method: "HEAD",
-                headers: {
-                    "User-Agent": "Mozilla/5.0 (compatible; NextAIDrawio/1.0)",
-                },
+                headers: { "User-Agent": USER_AGENT },
                 signal: headController.signal,
             })
-            clearTimeout(headTimeout)
             const contentType = headResponse.headers.get("content-type")
             if (contentType?.includes("application/pdf")) {
                 return NextResponse.json(
@@ -60,6 +57,8 @@ export async function POST(req: Request) {
                 "HEAD pre-check failed, proceeding with extraction:",
                 err,
             )
+        } finally {
+            clearTimeout(headTimeout)
         }
 
         // Extract article content with timeout to avoid tying up server resources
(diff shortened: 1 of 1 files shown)


ElshadHu:
> @ElshadHu I forgot to mention in issue #693 that we need to support i18n in error message. Could you please add this? Also, please check the Copilot messages. Thank you.

Sure, I will take a look at them.

ElshadHu:
Hey @Biki-dev 
The API routes don't use i18n.  Do you have any approach in mind?
Also, I would like to confirm, should I add a new test for pdf url or is this sufficient?

Biki-dev:
Hey @ElshadHu,

Thanks for the clarification about i18n in API routes. You're absolutely right that API routes don't use i18n directly. Let me address both parts of your question:

## **About the i18n approach:**

**Sorry for the confusion - upon checking the codebase more carefully, I noticed that all other similar error messages in the API routes aren't i18n-supported either.** 

Looking at the existing patterns, error messages from API routes are returned as simple English strings, and there's no established pattern for i18n support in API responses throughout the codebase.

Let's skip the i18n feature for this PR to maintain consistency with the existing code patterns. We can leave this as a simple English error message, just like all other API errors in the codebase. The i18n implementation for API responses could be considered as a future enhancement if needed.

*(Possible i18n solution for future reference: Return error codes like "PDF_URL_NOT_SUPPORTED" from API, then map to translations in client-side catch blocks)*

## **About testing:**

Your current test is **sufficient** for this fix.

Thanks for pointing this out!

ElshadHu:
@Biki-dev I appreciate for feedback 🙌

PR #710, opened 2026-02-26
fix: Changes to make npm run check GH status check pass

bnevis-i:
The Lint & Unit Tests PR status check is failing at the "Run lint" step over a half-dozen issues. This is causing all PR's to fail the Lint & Unit tets check. This fix resolves those issues.


DayuanJiang:
@bnevis-i  Thanks again.

PR #713, opened 2026-02-27
feat: enable save button in editor and update default filename format

Jay-ju:


DayuanJiang:
Thanks for the contribution! Unfortunately, I'll need to close this PR.
                                                                                                
The native draw.io save button was intentionally removed in #507 because it caused a race condition — after clicking 'Save' in the save dialog, the dialog would reappear due to a conflict between draw.io's native save events and our custom save logic.

Users can still save using our custom save button which supports multiple formats (drawio/png/svg).

PR #715, opened 2026-02-27
fix: biome schema check errors

bnevis-i:
Stronger fix for commit dd9d79d2d67a47fdd5ca9d72cdf3aa4f1103f956

Update the carat version of biome to ^2.4.4
(which is currently what is installed with ^2.3.10) and switch the biome.json to use the local configuration schema

This is because npm will take every minor version bump, but config schema will never be able to catch up.
Instead, point at the local config schema version, so that these are always in sync without
locking biome to a specific major/minor/patch release.

Specific error that is fixed when "biome ci" is invoked:

```
  ℹ The configuration schema version does not match the CLI version (local installed version)

  > 2 │     "$schema": "https://biomejs.dev/schemas/2.3.14/schema.json",
      │                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://biomejs.dev/schemas/2.3.14/schema.json",
+    "$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
     "vcs": {
         "enabled": true,
         "clientKind": "git",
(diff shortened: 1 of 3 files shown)


bnevis-i:
Revoking PR as it violates the best practices at https://biomejs.dev/internals/versioning/.  Will resubmit.

PR #716, opened 2026-02-27
fix: Eliminate spurious biome schema version errors on "biome ci" command

bnevis-i:
Stronger fix for commit dd9d79d2d67a47fdd5ca9d72cdf3aa4f1103f956

Apply the BKM at https://biomejs.dev/internals/versioning/ that recommends pinning a specific version of biome (current version of package-lock.json uses 2.4.4, which was previously installed with ^2.3.10)

Otherwise, if npm is allowed to upgrade at its discretion, project will have to continually chase the latest schema version
(as demonstrated in previous attempted fix).

The specific error that is fixed when "biome ci" is invoked:

```
  ℹ The configuration schema version does not match the CLI version (local installed version)

  > 2 │     "$schema": "https://biomejs.dev/schemas/2.3.14/schema.json",
      │                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

package-lock.json is also updated to reflect the package.json change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12245,7 +12245,6 @@
             "integrity": "sha512-glMJgnTreo8CFINujtAhCgN96QAqApDMZ8Vl1r8f0QT8QprvC1UCltV4CcWj20YoIyLZx6IUskaJZ0NV8fokcg==",
             "dev": true,
             "license": "MIT",
-            "peer": true,
             "dependencies": {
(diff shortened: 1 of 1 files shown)


DayuanJiang:
Thanks, removed some unnecessary changes to package-lock.json.

PR #738, opened 2026-03-18
Add  custom provider of models

nors1993:
**Add  custom provider of models  to use in private environment.**

image
image


DayuanJiang:
@nors1993  Thanks for the contribution!
Could you tell me in what situation we need this instead of using OpenAI provider with custom endpoint?

nors1993:
@DayuanJiang > @nors1993 Thanks for the contribution! Could you tell me in what situation we need this instead of using OpenAI provider with custom endpoint?

In private environment of organization, to use inner models.

DayuanJiang:
@nors1993 Thanks for the clarification. For this use case, you should be able to use the OpenAI provider with a custom base URL. Is there a specific reason that approach won't work for you?

nors1993:
> @nors1993 Thanks for the clarification. For this use case, you should be able to use the OpenAI provider with a custom base URL. Is there a specific reason that approach won't work for you?

@DayuanJiang   There are  not only OpenAI compatible APIs.

DayuanJiang:
@nors1993  Sorry but I still can't understand in what situation this pr is needed, could you give me a concrete example?

PR #741, opened 2026-03-20
feat: add glm vision model check

sbilly:
Add checks for GLM text and visual model naming conventions.

### DIFF
--- a/lib/ai-providers.ts
+++ b/lib/ai-providers.ts
@@ -1335,6 +1335,16 @@ export function supportsImageInput(modelId: string): boolean {
         return false
     }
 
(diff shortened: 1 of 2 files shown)


DayuanJiang:
@sbilly 
Thanks for the PR! 

I've pushed two additional commits:                                                         
1. Simplified the GLM vision detection — removed the redundant includes("v-") check (could false-positive on  names like "glm-dev-*") and updated comments with real model names.
2. Added vision detection for MiniMax (text-only) and Moonshot v1 (text-only), and fixed qwen3.5-flash being  incorrectly blocked (it's a vision model).

Issue #745, opened 2026-03-20
[Feature] Add Editable SVG (xmlsvg) Export Option


DayuanJiang:
## Feature Description

Add an "Editable SVG" option to the save dialog. This format embeds draw.io XML inside the SVG, making it both previewable and re-editable.

Related: #740

## Current Behavior

Save dialog offers 3 formats: `.drawio`, `.png`, `.svg` (view-only). The `xmlsvg` format is used internally but not exposed to users.

## Proposed Solution

Add `xmlsvg` as a 4th save format. Files to change:

- **`components/save-dialog.tsx`** — Add `"xmlsvg"` to `ExportFormat` type and `FORMAT_OPTIONS`
- **`contexts/diagram-context.tsx`** — Handle `xmlsvg` in `saveDiagramToFile`: decode the base64 data URL and save as `.svg`
- **`lib/i18n/dictionaries/*.json`** — Add label strings (e.g. `"xmlsvg": "Editable SVG"`)

No new dependencies needed. The `xmlsvg` format is already used throughout the codebase internally.

PR #761, opened 2026-03-26
feat: add xmlsvg export option

zongxi1115:
## Desc
Add an "Editable SVG" (xmlsvg) option to the save dialog. This format embeds draw.io XML inside the SVG, making it both previewable and re-editable in draw.io.

related to #745 

## Changes

   - `components/save-dialog.tsx`: Added xmlsvg to ExportFormat type and FORMAT_OPTIONS array
   - `contexts/diagram-context.tsx`: Added handling for xmlsvg export - decodes base64 data URL and saves as  .drawio.svg
     file
   - `lib/i18n/dictionaries/*.json`: Added label strings for all 4 languages

### DIFF
--- a/contexts/diagram-context.tsx
+++ b/contexts/diagram-context.tsx
@@ -314,17 +314,8 @@ export function DiagramProvider({ children }: { children: React.ReactNode }) {
                     mimeType = "image/png"
                     extension = ".png"
                 } else if (format === "xmlsvg") {
-                    // Editable SVG: decode base64 data URL and save as SVG
-                    let svgContent = exportData
-                    if (svgContent.startsWith("data:image/svg+xml;base64,")) {
-                        // Decode base64 to get the actual SVG content
-                        const base64Data = svgContent.replace(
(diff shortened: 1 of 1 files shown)


Zevan770:
Is it better to use `.drawio.svg` as file extension?

zongxi1115:
> Is it better to use `.drawio.svg` as file extension?

Indeed,I've changed it,thx!

DayuanJiang:
@zongxi1115  @Biki-dev 
Thanks for the PR! The feature looks good overall.
I found one issue with the atob() decode path — it corrupts non-ASCII text (Chinese, Japanese, emoji) due to UTF-8 double-encoding when creating the Blob. For example, `你好世界` becomes `ä½ å¥½ä¸ç` in the exported file.

I pushed a fix (e8be0a9) that simplifies the xmlsvg export by passing the data URL directly to the download handler, the same way PNG export already works. This avoids the atob() issue entirely.

PR #762, opened 2026-03-27
feat: show token usage and user-configurable estimated cost

Zhangxiaoyi688:
## Summary
- show input, output, and total token usage after each diagram generation
- add per-model pricing inputs in the model configuration dialog, stored locally in the browser
- calculate estimated cost from the user-configured per-1M token prices instead of server-side pricing config

## Details
- pricing is configured per user model in the UI
- cache read/write prices are optional and fall back to the input price when left blank
- if no pricing is configured, the app still shows token usage and marks cost as unavailable

## Validation
- npm run check
- npm run test -- --run
- npx playwright test tests/e2e/diagram-generation.spec.ts


Biki-dev (commented):
Great work on this PR overall. The feature is valuable and the implementation is clean end-to-end.

## What’s good

- Clear metadata model (`ChatUsageMetadata` / `ChatMessageMetadata`) and good type flow from API to UI.
- Pricing parsing + cost calculation logic is straightforward and easy to follow.
- UI usage summary is readable and useful for users.
- Test coverage was added for core pricing behavior and session metadata persistence.

## Two robustness fixes I recommend before merge

### 1) Tighten provider validation (avoid inherited object keys)

Current `ProviderNameSchema` validation using `val in PROVIDER_INFO` can accidentally accept inherited keys (for example `toString`).  
Use own-property checking instead.

**Suggested update** in `lib/server-model-config.ts`:

```ts
export const ProviderNameSchema: z.ZodType<ProviderName> = z
    .string()
    .refine((val): val is ProviderName => Object.hasOwn(PROVIDER_INFO, val), {
        message: "Invalid provider name",
    })
```

### 2) Guard metadata clone in session sanitizer

`sanitizeMessage` currently clones unknown persisted `metadata` without try/catch.  
If legacy/corrupted data includes non-cloneable values, this can throw and break session restore.

**Suggested update** in `lib/session-storage.ts`:

```ts
let metadata: ChatMessageMetadata | undefined
if (msg.metadata && typeof msg.metadata === "object") {
    try {
        metadata = structuredClone(msg.metadata as ChatMessageMetadata)
    } catch {
        // Drop invalid legacy metadata instead of failing restore
        metadata = undefined
    }
}

return {
    id: msg.id as string,
    role: role as "user" | "assistant" | "system",
    metadata,
    parts,
}
```

## Regression tests to add

### A) Provider schema rejects prototype keys

In `tests/unit/server-model-config.test.ts` add a case like:

```ts
it("rejects inherited object keys as provider names", () => {
    const invalidConfig = {
        providers: [
            {
                name: "Prototype Provider",
                provider: "toString" as any,
                models: ["model-1"],
            },
        ],
    }

    expect(() => ServerModelsConfigSchema.parse(invalidConfig as any)).toThrow()
})
```

### B) Sanitizer handles non-cloneable metadata safely

In `tests/unit/session-storage.test.ts` add a case like:

```ts
it("drops non-cloneable metadata instead of throwing", () => {
    expect(() =>
        sanitizeMessage({
            id: "assistant-2",
            role: "assistant",
            metadata: { bad: () => "not cloneable" },
            parts: [{ type: "text", text: "Done" }],
        }),
    ).not.toThrow()

    const sanitized = sanitizeMessage({
        id: "assistant-2",
        role: "assistant",
        metadata: { bad: () => "not cloneable" },
        parts: [{ type: "text", text: "Done" }],
    })

    expect(sanitized?.metadata).toBeUndefined()
})
```

These are small changes but they improve safety for real-world persisted data and close edge-case validation gaps.

Biki-dev (approved):

PR #800, opened 2026-04-10
fix: allow Qwen3.5 models to use image input

### LINKED ISSUE #799: [Bug] Qwen3.5 多模态模型 提示不支持图片输入
本地VLLM部署的Qwen3.5，原生支持图片传入，其他平台也经过验证，可以图片传入。
但next-ai-draw-io连接后，添加了图片附件，提示：{"error":"The model "Qwen3.5" does not support image input. Please use a vision-capable model (e.g., GPT-4o, Claude, Gemini) or remove the image."}

A locally deployed Qwen3.5 via vLLM natively supports image input, and this capability has been verified on other platforms as well. However, after connecting to next-ai-draw-io and attaching an image, the following error message appears: {"error":"The model "Qwen3.5" does not support image input. Please use a vision-capable model (e.g., GPT-4o, Claude, Gemini) or remove the image."}

octo-patch:
Fixes #799

## Problem

When using a locally deployed Qwen3.5 model via vLLM (which natively supports image input), attaching an image in the chat returns:

```
{"error":"The model \"Qwen3.5\" does not support image input. Please use a vision-capable model (e.g., GPT-4o, Claude, Gemini) or remove the image."}
```

This is a false negative in `supportsImageInput()` in `lib/ai-providers.ts`. The function exempts `qwen3.5-plus` and `qwen3.5-flash` as vision-capable, but not the base `qwen3.5` model name.

## Solution

Simplify the Qwen exception from two separate suffix checks to a single `qwen3.5` substring check. Since `"qwen3.5"` is a common prefix of all three variants (`qwen3.5`, `qwen3.5-plus`, `qwen3.5-flash`), this covers all of them with less code.

```diff
-    !lowerModelId.includes("qwen3.5-plus") &&
-    !lowerModelId.includes("qwen3.5-flash")
+    !lowerModelId.includes("qwen3.5")
```

Older Qwen text-only models (`qwen-turbo`, `qwen-plus`, `qwen3-max`, etc.) are still correctly blocked.

## Testing

- Added `supportsImageInput("Qwen3.5")` and `supportsImageInput("qwen3.5")` to the existing unit test suite.
- Existing tests for `qwen3.5-plus`, `qwen3.5-flash`, text-only Qwen models, and vision-indicator models remain passing.

### DIFF
--- a/lib/ai-providers.ts
+++ b/lib/ai-providers.ts
@@ -1356,12 +1356,11 @@ export function supportsImageInput(modelId: string): boolean {
     }
 
     // Qwen text models (not vision variants like qwen-vl)
-    // qwen3.5-plus is a vision model
+    // Qwen3.5 series (qwen3.5, qwen3.5-plus, qwen3.5-flash) natively support image input
     if (
         lowerModelId.includes("qwen") &&
         !hasVisionIndicator &&
-        !lowerModelId.includes("qwen3.5-plus") &&
-        !lowerModelId.includes("qwen3.5-flash")
+        !lowerModelId.includes("qwen3.5")
     ) {
         return false
     }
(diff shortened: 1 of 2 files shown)


DayuanJiang:
Splendid! Thanks for the quick fix.

PR #816, opened 2026-04-16
fix: always re-fetch access code config when settings dialog opens

### LINKED ISSUE #811: [Bug] docker部署配置了ACCESS_CODE_LIST的情况下对话框输入弹出设置框，但是没有输入密码的地方
> **Note**: This template is just a guide. Feel free to ignore the format entirely - any feedback is welcome! Don't let the template stop you from sharing your thoughts.

## Bug Description
配置了ACCESS_CODE_LIST的情况下对话框输入弹出设置框，但是没有输入密码的地方

## Steps to Reproduce
1. 配置了ACCESS_CODE_LIST
2. 启用对话功能
3. 弹出设置框
4. 没有要输入密码的地方

## Expected Behavior
输入密码

## Actual Behavior
What actually happened.

## Screenshots

Image

## Environment
- OS: [e.g. Windows 11, macOS 14]
- Browser: [e.g. Chrome 120, Safari 17]
- Version: [e.g. 1.0.0]

## Additional Context
Any other information about the problem.

octo-patch:
Fixes #811

## Problem

When `ACCESS_CODE_LIST` is configured on the server, the settings dialog was not showing the access code input field in two cases:

1. **Stale localStorage cache**: if a user had previously visited the app without `ACCESS_CODE_LIST` enabled, the cached value (`accessCodeRequired=false`) was used indefinitely, permanently hiding the password input field even after the admin enabled access code protection.

2. **Race condition on first visit**: the settings dialog could be opened (e.g. triggered automatically by an authentication error) before the async fetch to `/api/config` completed, causing the dialog to render without the access code input field.

## Solution

Replace the mount-only fetch (guarded by a cache check) with a fetch that runs whenever the dialog opens (`open` added to the dependency array). This ensures the settings dialog always reflects the server's current `ACCESS_CODE_LIST` configuration.

The localStorage cache is still updated on each successful fetch, so the initial render can still use the cached value for a fast display—but the dialog immediately corrects itself with fresh data from the server.

## Testing

1. Deploy with `ACCESS_CODE_LIST` set
2. Open the settings dialog — the access code input should appear
3. Clear localStorage, revisit, then open settings — input still appears
4. Deploy without `ACCESS_CODE_LIST`, visit once (caches `false`), then redeploy with `ACCESS_CODE_LIST` set, reopen settings — input now correctly appears instead of being hidden by the stale cache

### DIFF
--- a/components/settings-dialog.tsx
+++ b/components/settings-dialog.tsx
@@ -134,8 +134,11 @@ function SettingsContent({
     const [isApplyingProxy, setIsApplyingProxy] = useState(false)
 
     useEffect(() => {
-        // Only fetch if not cached in localStorage
-        if (getStoredAccessCodeRequired() !== null) return
+        // Re-fetch config whenever the dialog opens to ensure we always show
+        // the access code input if the server requires it. This fixes the case
+        // where a stale localStorage cache (from before ACCESS_CODE_LIST was
+        // configured) would hide the access code input.
+        if (!open) return
 
         fetch(getApiEndpoint("/api/config"))
             .then((res) => {
@@ -151,10 +154,9 @@ function SettingsContent({
                 setAccessCodeRequired(required)
             })
             .catch(() => {
-                // Don't cache on error - allow retry on next mount
-                setAccessCodeRequired(false)
+                // Keep existing cached value on error
             })
-    }, [])
+    }, [open])
 
     // Detect current language from pathname
     useEffect(() => {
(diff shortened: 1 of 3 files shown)


Biki-dev:
@octo-patch `fix: allow QvQ (Qwen Visual QA) models to use image input`
Is this commit unrelated to the PR? Please keep only the relevant changes.

DayuanJiang:
@octo-patch  Thanks for the contribution, I have removed the unrelated code.

PR #818, opened 2026-04-20
fix: add min-h-0 to sidebar to fix scroll constraint

### LINKED ISSUE #576: [Bug] Add model button becomes inaccessible when adding 6+ models to left-side provider (scrolling restricted)
**Describe the bug**
When configuring models in the left-side "AI Model Configuration" panel, adding around 6 models pushes the "Add model" button down out of view. The panel's scrolling is restricted, so the button cannot be scrolled into view or selected.

**Steps to reproduce**
1. Open the **AI Model Configuration** modal.
2. Select a provider on the left (e.g. *Google Vertex AI*).
3. Add models repeatedly.
4. Continue until you have ~6 models listed in the provider config.
5. Observe that the "Select model" control moves below the visible area and is not reachable because the left-side panel does not allow scrolling to that button.

**Observed behavior**
- The "Select model" button is pushed down and becomes inaccessible.
- The user cannot scroll far enough to select the button.
- This prevents adding more models or interacting with controls at the bottom of the panel.

**Expected behavior**
- The left-side configuration panel should allow scrolling so all controls remain reachable.

**Screenshots**

Image

LaaraibAhmed:
### Fixes #576 

The "Add model" button becomes inaccessible when 6+ models are added to a provider in the AI Model Configuration modal.

# Problem
The left-side provider list panel uses flex-1 on the ScrollArea component to fill available height, but the parent sidebar div was missing min-h-0. In CSS flexbox, elements default to min-height: auto which allows them to grow beyond their parent's bounds — this prevented the height constraint from propagating down, so ScrollArea had nothing to scroll against and the content simply overflowed out of view instead of becoming scrollable.

# Solution
Added min-h-0 to the sidebar container div in model-config-dialog.tsx:

Before:
```tsx
<ScrollArea className="flex-1 px-2">
```
After:
```tsx
<ScrollArea className="flex-1 px-2 min-h-0">
```
This allows the flex height constraint from the parent dialog to propagate correctly into the sidebar, giving ScrollArea a bounded height to scroll within. All controls remain reachable regardless of how many models are added.

# Changes
components/model-config-dialog.tsx — added min-h-0 to sidebar container (1 line change)

### DIFF
--- a/components/model-config-dialog.tsx
+++ b/components/model-config-dialog.tsx
@@ -405,7 +405,7 @@ export function ModelConfigDialog({
                             </span>
                         </div>
 
-                        <ScrollArea className="flex-1 px-2">
+                        <ScrollArea className="flex-1 px-2 min-h-0">
                             <div className="space-y-1 pb-2">
                                 {config.providers.length === 0 ? (
                                     <div className="px-3 py-8 text-center">


Biki-dev (changes requested):
Hi, thanks for the PR. However, `package.json` and the new `bun.lock` file are not related to this PR. Please remove them. Thank you.

LaaraibAhmed:
Hi Bikidev,
Thanks for reviewing my pr. Will do so today.

LaaraibAhmed:
Hi,
I have removed the files as requested.

Biki-dev:
Hi @LaaraibAhmed, As I mentioned, “remove it” doesn’t mean deleting it from the codebase—I just meant removing it from this PR.

I only want changes related to components/model-config-dialog.tsx in this PR. If you’re having difficulty doing that, you can close this PR and open a new one with only the one-line change in that file. Alternatively, you can fix it in this PR—either option works for me.

Thank you.

Biki-dev:
Your mistakes:

1. In your first push (before my initial review), you added a new file `bun.lock`, which is not required in the codebase. You also made changes to `package.json` and `.gitignore`, which were not necessary.

2. In your most recent commit, after I asked you to remove those changes and delete `bun.lock` entirely (since we are using npm and not Bun), you made another mistake. While deleting `bun.lock` was correct, you also accidentally deleted `package-lock.json` and made unnecessary changes to `.gitignore` and `package.json`. It seems like instead of simply removing `bun.lock`, you attempted to ignore it via `.gitignore`, which is not the correct approach.

The correct approach:

This PR is only meant to fix a CSS/UI issue, so it should contain changes to **only one file**:
`components/model-config-dialog.tsx`

No other files should be modified or included in the PR. Only the relevant changes to this file should appear in the PR’s change list.

LaaraibAhmed:
Hi @Biki-dev
Thanks for this. Made the changes.

Biki-dev (approved):

DayuanJiang:
@LaaraibAhmed  @Biki-dev 
Thanks!

PR #835, opened 2026-05-15
feat: add all Draw.io themes to settings panel

ruguoba:
Closes #499

## Changes
- Add all 6 available Draw.io themes to the settings panel dropdown
- Themes: kennedy, atlas, dark, min, sketch, simple
- Replace the toggle button with a Select dropdown for better UX

## Context
The settings panel previously only offered 2 of the 6 available Draw.io themes (min and sketch) via a toggle button. This PR expands the dropdown to include all themes, giving users full control over the editor appearance.

### Files modified
- `app/[lang]/page.tsx` - Updated state type and handler to support all 6 themes
- `components/settings-dialog.tsx` - Replaced toggle button with dropdown selector
- `components/chat-panel.tsx` - Updated prop types for theme selection

Ref: https://www.drawio.com/doc/faq/configure-themes

PR #837, opened 2026-05-16
fix(e2e): resolve iframe toolbar strict mode violation

ruguoba:
## Problem

CI E2E test `can interact with draw.io toolbar` is failing with:

```
Error: strict mode violation: locator('text="Diagram"').or(locator('[title*="Diagram"]')) resolved to 2 elements
```

The selector matches two different elements:
1. `<div title="Diagram" class="geFormatTitle">` (via `[title*="Diagram"]`)
2. `<div>Diagram</div>` (via `text="Diagram"`)

## Solution

Use `.first()` with the more specific `[title*="Diagram"]` selector to avoid matching multiple elements.

## Changes

- `tests/e2e/iframe-interaction.spec.ts`: Fixed selector in "can interact with draw.io toolbar" test

## Testing

This fix resolves the strict mode violation while still verifying that the draw.io toolbar is accessible.

Fixes the E2E test failure in CI: https://github.com/DayuanJiang/next-ai-draw-io/actions/runs/25922452973/job/76194926866


ruguoba:
This PR fixes the strict mode violation introduced after #835. I verified that the failing test still fails on main and passes on this branch. Squash merge is preferred to keep the history clean.

DayuanJiang:
@ruguoba  Thanks, I rebased this to the main so the change is more clean.

PR #857, opened 2026-06-01
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade the MiniMax model lineup to include the latest M3 model as the default suggestion.

## Changes
- Add `MiniMax-M3` to the suggested model list (set as new default, listed first)
- Retain `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` as available alternatives
- Remove deprecated `MiniMax-M2.5` / `MiniMax-M2.5-highspeed` from the suggested list
- Update `supportsImageInput`: M2.x are still text-only, M3 supports image input
- Update unit tests to reflect the new lineup (text-only assertions for M2 series, image-input assertion for M3)
- Update `AI_MODEL` example in the CN / EN / JA AI provider docs to `MiniMax-M3`

## Why
`MiniMax-M3` is the new flagship MiniMax model with a 512K context window, 128K max output, and image-input support across both OpenAI- and Anthropic-compatible interfaces. Older M2.5 / M2.5-highspeed have been superseded; M2.7 / M2.7-highspeed are kept so existing users still see their current default.

## Testing
- `npx vitest run` — 105/105 unit tests pass
- `npx tsc --noEmit` — clean
- `npx biome check` — no new errors on changed files

### DIFF
--- a/docs/cn/ai-providers.md
+++ b/docs/cn/ai-providers.md
@@ -215,7 +215,7 @@ MiniMax 支持两种 API 格式：
 
 ```bash
 MINIMAX_API_KEY=your_api_key
-AI_MODEL=MiniMax-M2.7
+AI_MODEL=MiniMax-M3
 ```
 
 可选配置：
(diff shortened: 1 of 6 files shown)


DayuanJiang:
Thanks for the contribution!

PR #858, opened 2026-06-03
fix: block private IPv6 URLs

chaochaoweb3:
## Summary
- normalize bracketed IPv6 hostnames before SSRF checks
- block IPv6 loopback, unique-local, link-local, and IPv4-mapped hosts
- add regression coverage for IPv6 private URL bypasses while preserving public hostnames

## Verification
- RED: `npx vitest run tests/unit/ssrf-protection.test.ts` failed with `expected false to be true` for `http://[::1]/`
- GREEN: `npx vitest run tests/unit/ssrf-protection.test.ts` passed
- `npm test -- --run` passed, 8 files / 107 tests
- pre-push `npm test -- --run` passed, 8 files / 107 tests
- `npx tsc --noEmit` passed
- `npm run check` exited 0; it reported existing Biome schema/version info and unrelated warnings
- `git diff --check` passed

### DIFF
--- a/lib/ssrf-protection.ts
+++ b/lib/ssrf-protection.ts
@@ -11,7 +11,10 @@ export function isPrivateUrl(urlString: string): boolean {
         const url = new URL(urlString)
         // Strip a trailing dot so FQDN forms like "localhost." (which still
         // resolve to 127.0.0.1) cannot bypass the equality checks below.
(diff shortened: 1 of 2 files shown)


DayuanJiang:
@chaochaoweb3 Thanks for the contribution, I have fixed some minor issue.

chaochaoweb3:
Noted the minor follow-up fix on merge. No further action from my side.

PR #862, opened 2026-06-07
feat(mcp): add multi-page (mxfile) support to MCP server

Siddhant0507Shekhar:
## Summary

The MCP server can already _read_ multi-page draw.io documents (the embedded editor and `get_diagram` both round-trip the full `<mxfile>` shape), but the _write_ path was single-page only. Passing a multi-page payload to `create_new_diagram` failed validation with the misleading message **"Expected closing tag `</root>` but found `</mxCell>`"** because the validator assumed input was a bare `<mxGraphModel>` and could not walk past the `<mxfile><diagram>...</diagram></mxfile>` wrapper.

This PR closes the gap end-to-end. The MCP surface now supports the natural drawio multi-page workflow: an agent can `add_page` a CNN diagram alongside an existing Transformer diagram, `edit_diagram` either page by id/name/index, and export specific pages — all while keeping every existing single-page caller working unchanged.

## Motivating user scenario

```
User: "create a different page in same session of draw io, and in that page I want a CNN diagram"
```

Today the only workaround is packing both diagrams onto a single canvas with a visual divider — losing tab navigation, forcing a shared coordinate space, and scaling poorly past 2 diagrams. After this PR the workflow is one tool call per page.

## What's in this PR

**New helper module** — `packages/mcp-server/src/pages.ts` (~250 LOC)
Centralises page CRUD so every layer agrees the canonical in-memory shape is always `<mxfile>`. Exports: `normalizeToMxfile`, `parseMxfile`, `serializeMxfile`, `listPagesFromDoc`, `findPageElement`, `addPageToDoc`, `renamePageInDoc`, `deletePageFromDoc`, `generatePageId`, plus `PageSelector` / `PageInfo` types and `hasPageSelector`.

**Validator made page-aware** — `packages/mcp-server/src/xml-validation.ts`
- `checkDuplicateIds` now detects an `<mxfile>` root and scopes the cell-id uniqueness check per `<diagram>`. The legacy regex check would otherwise reject every multi-page doc because cells `"0"` and `"1"` repeat in each page's `<root>` by design.
- Adds a `<diagram>` id uniqueness check across the document.
- `autoFixXml` step 22 (duplicate-id rename) is now guarded against mxfile inputs — silently renaming the sentinel root cells would corrupt drawio's `parent` references.

**Ops layer page-scoped** — `packages/mcp-server/src/diagram-operations.ts`
`applyDiagramOperations` accepts an optional `PageSelector`. For `<mxfile>` input it resolves the target page first and scopes all `querySelectorAll` calls to that page's `<root>`. So a `delete` on page 2's cell `"2"` no longer touches page 1's cell `"2"`. Legacy bare-`<mxGraphModel>` callers continue to work; passing a selector against a non-multi-page doc returns a clear error.

**Tool surface** — `packages/mcp-server/src/index.ts`

| Tool | Change |
|---|---|
| `create_new_diagram` | Accepts either a bare `<mxGraphModel>` (legacy, auto-wrapped) **or** a full `<mxfile>` with N `<diagram>` children |
| `edit_diagram` | New optional `page_id` / `page_name` / `page_index` (defaults to first page) |
| `get_diagram` | New optional page selector; with no selector returns the full mxfile, with selector returns a single-page projection |
| `export_diagram` | `.drawio` with selector writes a single-page mxfile; PNG/SVG with selector switches the editor tab via `selectPage` postMessage before exporting |
| `list_pages` | **new** — returns `[{id, name, index, cellCount}]` |
| `add_page` | **new** — appends a page without touching existing tabs or unsaved user edits |
| `rename_page` | **new** |
| `delete_page` | **new** — refuses to delete the last remaining page |

All tool description strings were updated to teach the LLM client the multi-page semantics, including a sharper warning on `create_new_diagram` that it now wipes _every_ tab (use `add_page` for additive operations).

**Browser bridge** — `packages/mcp-server/src/http-server.ts`
Added `pendingPageSelectId` to `SessionState` and the polling response. When the MCP server requests a PNG/SVG export of a specific page, the iframe bridge sends drawio `{action:'selectPage', pageId}` and waits 400 ms for the tab switch to settle before triggering the export. The wrapper-injection hack at the `.drawio` save path is now redundant since session XML is canonically mxfile — kept as a defence-in-depth fallback with an explanatory comment.

Screenshot 2026-06-07 at 12 49 52 PM
Screenshot 2026-06-07 at 12 49 59 PM

## Backward compatibility

Every existing single-page caller keeps working without modification:
- `create_new_diagram` with a bare `<mxGraphModel>` is auto-wrapped into a single-page `<mxfile>`. Output `get_diagram` shape is unchanged for these callers (a one-`<diagram>` mxfile, which is what the read path already returned).
- All page-selector parameters are optional. When omitted on a single-page document, behaviour is bit-identical to before.
- No tool was renamed or removed.

## Testing

**New unit tests** — `packages/mcp-server/tests/multi-page.test.ts` (34 tests):
- backward compat: bare `<mxGraphModel>` still validates and round-trips
- validator: accepts a valid two-page mxfile; rejects in-page duplicate cell ids; rejects duplicate `<diagram>` ids; preserves root sentinel `0`/`1` cells during autofix
- ops: page-targeted add/update/delete; defaults to first page; clear error when target page is missing; cells on other pages are unaffected
- pages helpers: list, find by id/name/index with correct precedence, add, rename, delete, refuse-last-page
- end-to-end: the motivating Transformer + CNN scenario reproduces with two tool-equivalent calls

**New smoke test** — `packages/mcp-server/tests/smoke.mjs`:
Spawns the built `dist/index.js`, performs the MCP `initialize` handshake, calls `tools/list`, asserts all 9 tools are present and that `edit_diagram` / `add_page` advertise the new page-selector schema fields. Catches wiring regressions independent of unit tests.

**Suite results locally:**
- `npm run test` at root: **107 / 107 pass** (no regressions in existing app code)
- `npm run test` in `packages/mcp-server`: **34 / 34 pass**
- `node tests/smoke.mjs` in `packages/mcp-server`: pass
- Biome `check` clean (errors=0; remaining warnings are `noNonNullAssertion` in tests, common pattern in vitest helpers)
- `tsc --noEmit` clean

## How to verify locally

```bash
cd packages/mcp-server
npm install
npm run build
npm test                        # 34 unit tests
node tests/smoke.mjs            # MCP JSON-RPC smoke test
```

Then point your MCP client (Claude Desktop / Cursor / Claude Code) at `node /absolute/path/to/dist/index.js`, call `start_session`, and try:
1. `create_new_diagram` with a two-page `<mxfile>` — both tabs render
2. Or single-page → `add_page (name="CNN")` → `edit_diagram (page_name="CNN", operations=[...])`

## Notes for review

- **CONTRIBUTING.md** recommends opening an issue first for significant changes. Apologies for going straight to a PR — happy to convert this into an issue + design discussion if preferred, or to split the change into smaller PRs (helpers → validator → ops → tools) if that's easier to review. The `packages/mcp-server/src/pages.ts` module is a natural seam to land first.
- **Package-local Vitest**: a `vitest.config.ts` and `tests/` directory were added to `packages/mcp-server/` because the root vitest is configured with `environment: "jsdom"` for the Next.js app, and the MCP server's DOM polyfill needs `environment: "node"`. CI's existing root-level test step is untouched; package tests are runnable via `npm test` inside the package.
- **AI-assisted**: per the CONTRIBUTING "Using AI Tools" section — every line was reviewed before this PR. The plan was scoped against the actual code (file:line citations) before implementation, and the implementation was unit-tested before the smoke test, before the build, before this PR.


DayuanJiang:
Thanks for the contribution! I'd been wanting to implement this feature but couldn't figure out how.

One request: Could you remove "Claude" from the contributors? AI assistance is fine, but it doesn't need to appear as a co-author.

DayuanJiang:
Could you also check the review comments from Copilot? If they're valid, please fix them; if not, just click "Resolve conversation."

Siddhant0507Shekhar:
Thanks for the review @DayuanJiang. Pushed a new commit (`d6acb6f`) addressing everything you and Copilot raised.

DayuanJiang:
@Siddhant0507Shekhar  Thanks! I have fixed some minor issue. Merged.

PR #873, opened 2026-06-22
[codex] fix: block private IPv6 URLs in SSRF checks

chaochaoweb3:
## Summary

- normalize bracketed IPv6 hostnames before SSRF checks
- block IPv6 loopback/unspecified, unique-local, link-local, and IPv4-mapped private hosts
- add unit coverage for private IPv6 URLs and public hostname false positives

## Why

The existing SSRF guard handled private IPv4 and internal hostnames, but bracketed IPv6 literals could bypass those checks. This keeps the existing small guard and fills the IPv6 private-address gap.

## Validation

- `npm test -- --run tests/unit/ssrf-protection.test.ts`
- `npm test -- --run`
- `npm run check` exits 0; it still reports existing Biome warnings/schema info outside this diff
- `git diff --check`

### DIFF
--- a/lib/ssrf-protection.ts
+++ b/lib/ssrf-protection.ts
@@ -11,7 +11,10 @@ export function isPrivateUrl(urlString: string): boolean {
         const url = new URL(urlString)
         // Strip a trailing dot so FQDN forms like "localhost." (which still
         // resolve to 127.0.0.1) cannot bypass the equality checks below.
-        const hostname = url.hostname.toLowerCase().replace(/\.$/, "")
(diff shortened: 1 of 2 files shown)


chaochaoweb3:
Closing this because the branch now rebases to an empty diff against the latest `main`: the SSRF IPv6 coverage from this PR, including `::` and the full `fe80::/10` range, is already present upstream.

I verified the current upstream behavior locally with:
- `npm test -- tests/unit/ssrf-protection.test.ts --run`

Result: 1 file passed, 8 tests passed.

PR #881, opened 2026-07-01
fix: remove user-scalable restriction, clean up unused dep and memory leak

fix2015:
Found a few things worth cleaning up:

- the viewport config had `userScalable: false` which prevents zooming — that's a WCAG accessibility violation and makes things harder for anyone who needs to zoom in on mobile
- `base-64` was listed as a dependency but never imported anywhere — the code uses native `atob`/`btoa` instead
- `env.example` was missing about 9 supported providers from the options comment (edgeone, doubao, modelscope, glm, qwen, qiniu, kimi, minimax, sglang)
- there was a small memory leak in the validation retry tracking — if VLM validation threw an error, the retry count for that tool call ID was never cleaned up from the map

### DIFF
--- a/app/[lang]/layout.tsx
+++ b/app/[lang]/layout.tsx
@@ -25,8 +25,6 @@ const jetbrainsMono = JetBrains_Mono({
 export const viewport: Viewport = {
     width: "device-width",
     initialScale: 1,
-    maximumScale: 1,
-    userScalable: false,
 }
 
 // Generate static params for all locales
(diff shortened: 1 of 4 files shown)


Biki-dev (commented):
Thanks for the fixes.

However, for the `base-64` package, I can only see changes in `package.json`. Since `package-lock.json` is also part of the repository, it would be good to include the corresponding changes there as well.

Other than that, it's good to merge.

PR #884, opened 2026-07-02
fix: increase model validation token budget

### LINKED ISSUE #883: [Feature] Allow configuring max_tokens for model validation/testing
## Feature Description

Allow users to configure the `max_tokens` value used during model validation/testing for custom OpenAI-compatible providers.

## Problem Context

This is related to a false-negative model validation issue.

When using a custom OpenAI-compatible provider through LiteLLM, some reasoning/thinking models fail the model test even though they work correctly in normal requests.

The issue happens because the model test appears to send a very small `max_tokens` value, such as `20`. For reasoning models, the response may first contain reasoning tokens and only later produce the final `content`. With a small `max_tokens`, the request may end before any final content is generated, causing the provider/proxy to return an error like:

```text
litellm.BadGatewayError: upstream returned a successful but empty completion

For example, this kind of request may fail:

{
  "model": "glm-4.7",
  "messages": [
    {
      "role": "user",
      "content": "Only reply OK"
    }
  ],
  "max_tokens": 20,
  "stream": false
}

But the same model works when max_tokens is larger:

{
  "model": "glm-4.7",
  "messages": [
    {
      "role": "user",
      "content": "Only reply OK"
    }
  ],
  "max_tokens": 1000,
  "stream": false
}

So the model itself is usable, but the validation step marks it as unavailable.

Proposed Solution

Please make the model test/validation max_tokens configurable.

Possible options:

Add a setting for validation/test max_tokens.
Increase the default test max_tokens, for example to 1000.
Retry once with a larger max_tokens if the test fails with empty content.
Allow users to skip validation and save/use the model anyway.
Allow users to configure the test prompt and test parameters for custom providers.
Alternatives Considered

A workaround is to use a non-reasoning model, such as a flash/chat model, for validation. However, this prevents users from using reasoning models that are otherwise fully functional.

Another workaround is to modify the proxy or add another proxy layer to rewrite small max_tokens values, but that is not ideal and adds unnecessary complexity.

Additional Context

Environment:

Provider type: Custom OpenAI-compatible API
Proxy: LiteLLM
API endpoint: /v1/chat/completions
Example models:
glm-4.7
glm-5.1
glm-5.2
vendor-deepseek-v4-pro

The model works with a larger max_tokens, but fails during the application's model test because the output token limit is too small.

chaochaoweb3:
## Summary

- increase the model validation token budget from 20 to 1000
- apply the fixed budget to both ModelScope streaming validation and the general AI SDK validation request
- keep the change intentionally simple without adding configuration or admin settings

Fixes #883

## Validation

- npx biome check app/api/validate-model/route.ts
- npm test -- --run (153 tests passed)
- the pre-push hook reran the full unit suite successfully


DayuanJiang:
@chaochaoweb3 Thanks for the contribution!
I was trying to amend some issues in the PR, but after reviewing the code, I think it's better to just hardcode the max token to 1000 instead of adding a config option to do this. 
What do you think?

chaochaoweb3:
Agreed. I pushed 42ab388 to simplify this: both the ModelScope max_tokens value and the general maxOutputTokens value are now hardcoded to 1000, and the config, admin setting, translations, parser, and parser tests were removed. Validation: Biome passed for the route, and the full unit suite passed with 153 tests.

Issue #885, opened 2026-07-05
MCP Tool Error Response: Error: You must call get_diagram first before edit_diagram.


xulubao:
在kiro中画图的时候，经常性出现：
MCP Tool Error Response: Error: You must call get_diagram first before edit_diagram.

This ensures you have the latest diagram state including any manual edits the user made in the browser. Please call get_diagram, then use that XML to construct your edit operations.
的错误，我想问一下如何给提示词，我看kiro已经 get_diagram 之后才edit_diagram，但是还是提示失败。
kiro模型用的是 claude opus4.8

PR #890, opened 2026-07-11
fix(mcp): replace edit_diagram 30s time gate with content comparison

NgoQuocViet2001:
## Summary

- stop expiring a valid `get_diagram` snapshot after 30 seconds (the wall-clock gate rejected slow-reasoning clients even when they followed the documented workflow — #885)
- instead of removing staleness detection entirely, replace the timeout with a **content comparison**: the server remembers the exact state-store XML the model last saw (`lastSeenXml`, set by `get_diagram` / `create_new_diagram` / `edit_diagram` / page CRUD) and rejects `edit_diagram` only when the live browser state actually differs — i.e. the user made manual edits the model hasn't seen yet
- slow reasoning never trips the gate; unseen user edits still do, with a dedicated error message telling the model to re-fetch once and retry
- the gate compares the RAW browser-pushed XML on both sides (not the normalised mxfile form) so format differences can never be misread as content changes

Related to #885

## Validation

- MCP server TypeScript build passed
- `npm test`: 52/52 passed (45 existing + 4 new unit tests for the gate + wiring tests)
- Biome check passed
- Live end-to-end run against the real stdio server + browser:
  1. `edit_diagram` without `get_diagram` 12s after create (browser had loaded and pushed autosave) → **allowed**, no false positive
  2. simulated manual user edit (POST to `/api/state` like the browser autosave path), then `edit_diagram` without `get_diagram` → **rejected as stale**
  3. one `get_diagram`, then `edit_diagram` → **allowed**, recovery works
  4. `get_diagram`, wait 35s (> old 30s limit), `edit_diagram` → **allowed**, #885 scenario fixed

## Notes

Built on top of @NgoQuocViet2001's original diagnosis in this PR: the 30s wall-clock gate was the root cause of #885. The follow-up commit keeps that fix while restoring the ability to detect mid-session manual browser edits, which the one-time sentinel check gave up.

### DIFF
--- a/packages/mcp-server/package-lock.json
+++ b/packages/mcp-server/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@next-ai-drawio/mcp-server",
-    "version": "0.2.1",
+    "version": "0.2.2",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@next-ai-drawio/mcp-server",
-            "version": "0.2.1",
+            "version": "0.2.2",
             "license": "Apache-2.0",
             "dependencies": {
                 "@modelcontextprotocol/sdk": "^1.0.4",
--- a/packages/mcp-server/package.json
+++ b/packages/mcp-server/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@next-ai-drawio/mcp-server",
-    "version": "0.2.1",
+    "version": "0.2.2",
     "description": "MCP server for Next AI Draw.io - AI-powered diagram generation with real-time browser preview",
     "type": "module",
     "main": "dist/index.js",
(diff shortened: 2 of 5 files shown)


DayuanJiang:
Thanks for the PR, your diagnosis is spot on. However, the solution may run into trouble in the following scenario:
```
user: draw a cat
agent: create_diagram
user: change the eyes to blue
agent: get_diagram -> edit_diagram
(user: move the leg position in the panel)
user: make the leg shorter
agent: edit_diagram 
```
In the last step, the agent should have called get_diagram before editing, since the user had moved the leg position. But it didn't, because it had already used get_diagram earlier in the conversation.

I thought of a better way to solve this: since users rarely change the diagram themselves, the agent can just call get_diagram directly by default. I'd also add a mechanism to track whether the user changed the diagram. If they did, get_diagram will fail with a message: "please get_diagram before edit".

NgoQuocViet2001:
@DayuanJiang Thanks for the concrete scenario and the stronger design. Tracking actual diagram-content changes is much more robust than a conversation-level gate, and I appreciate you carrying it through to the merged implementation.

PR #894, opened 2026-07-13
fix: validate admin settings path before reporting writable

NgoQuocViet2001:
## Summary

- reject `SETTINGS_FILE` when it points to an existing non-file target
- keep parent-directory writability checks for new settings files
- replace the POSIX/root-permission-dependent test fixture with a cross-platform regression case
- keep the `0600` mode assertion on POSIX while still verifying persisted JSON on Windows

## Why

`isSettingsWritable()` only checked the target's parent directory. A configured settings path that was itself a directory was therefore reported as writable even though `saveSettings()` could not write the temporary file and rename it into place.

The existing negative test also used `/nonexistent-root-dir/settings.json`, which is not reliably unwritable on Windows, while POSIX mode bits are not meaningful in the same way on Windows. Together these assumptions caused two unit-test failures on a supported Electron development platform.

## Testing

- `npm run test -- --run` — 153 tests passed
- `npx tsc --noEmit`
- `npm run build`
- `npx biome check lib/admin/settings.ts tests/unit/admin-settings.test.ts`

### DIFF
--- a/lib/admin/settings.ts
+++ b/lib/admin/settings.ts
@@ -110,7 +110,12 @@ let writableCache: boolean | null = null
 export function isSettingsWritable(): boolean {
     if (writableCache !== null) return writableCache
     try {
-        const dir = path.dirname(getSettingsPath())
+        const filePath = getSettingsPath()
+        if (fs.existsSync(filePath) && !fs.statSync(filePath).isFile()) {
+            writableCache = false
+            return writableCache
+        }
+        const dir = path.dirname(filePath)
         fs.mkdirSync(dir, { recursive: true })
         fs.accessSync(dir, fs.constants.W_OK)
         writableCache = true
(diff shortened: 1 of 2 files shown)


NgoQuocViet2001:
Nudging this — still merges cleanly and all 7 checks are green.

Short version for triage: `isSettingsWritable()` only checked the *parent directory*, so a `SETTINGS_FILE` pointing at an existing directory was reported writable even though `saveSettings()` cannot write-then-rename onto it. The change validates the target itself while keeping the parent-directory check for files that do not exist yet.

It also replaces the negative test's `/nonexistent-root-dir/settings.json` fixture, which is not reliably unwritable on Windows, and scopes the `0600` mode assertion to POSIX — those two assumptions were failing the unit suite on a supported Electron dev platform.

Happy to rebase if it has drifted.

PR #895, opened 2026-07-13
fix: preserve multi-page diagrams after export

### LINKED ISSUE #879: [Bug]  关闭应用后部分历史记录消失
>
## Bug Description
在应用内，同一项目中多开页面。关闭应用后，再次打开，只保存了第一页的内容。其他也的，历史记录全部消失。并且希望加上能保存工程文件`.drawio`的选项。当前无法像drawio官方导出流程一样，先要求保存工程文件，再保存导出文件。以及是否有找回消失缓存的方法。

## Environment
- OS: [ Windows 11]
- Version: [e.g. v30.0.2]

NgoQuocViet2001:
## Summary
- Preserve the full diagram XML supplied by draw.io after XML/SVG exports instead of replacing state with the exported page only.
- Use the same full-document XML when downloading a `.drawio` file so every page is retained.
- Keep a compatibility fallback for export events that do not provide XML.
- Add focused regression coverage for multi-page preservation and fallback behavior.

## Problem
The export callback decoded `data.data`, which represents the currently exported page, and wrote that page-only XML back into `chartXML`. Exports used for AI context, history, thumbnails, or file saving could therefore shrink a multi-page document to one page before the next session persistence cycle.

`react-drawio` also exposes the source document as `EventExport.xml`. This change uses that full XML for persistent state and `.drawio` downloads while still returning the current-page extraction to existing AI/history callers.

Closes #879

## Verification
- `npx vitest run tests/unit/utils.test.ts` — 15/15 passed
- `npx vitest run --exclude tests/unit/admin-settings.test.ts` — 138/138 passed
- `npx tsc --noEmit`
- `npm run build`
- `npx biome check contexts/diagram-context.tsx lib/utils.ts tests/unit/utils.test.ts`
- `git diff --check`

## Windows baseline note
The complete root unit command passed 153/155 tests. Its two failures are the existing Windows assumptions in `tests/unit/admin-settings.test.ts` (unwritable POSIX root path and POSIX mode bits), already isolated in #894; this PR does not touch those files. Linux CI remains authoritative for the complete suite.


DayuanJiang:
Thanks for contribution. your diagnosis was correct. I pushed a follow-up commit  with a different fix strategy: EventExport.xml from xmlsvg exports contains compressed <diagram> payloads, which would break edit_diagram. Since autosave already provides the full document uncompressed, the export callback now just stops writing chartXML entirely. Your data.xml change for .drawio downloads is kept.

